PR #37, opened 2020-06-28
Update PHP CS Docs per Revision 63943

ntwb:
This PR updates the CS PHP docs from the latest revision made 16 July 2019

> https://make.wordpress.org/core/wp-admin/revision.php?revision=63943&gutenberg=true


ntwb:
Yeah, I agree @GaryJones, now that everything from the PHP handbooks is in, the examples can begin to be updated in a more collaborative way via PRs here 😃

Thanks for the 👍🏼's and merge y'all

PR #40, opened 2020-07-22
Update css.md - amend sample of incorrect values

secan:
wp_css-md_screenshot

In line 9-13 of the sample it looks like the text formatting is incorrect and the information about not using unit for line height is missing

### DIFF
--- a/wordpress-coding-standards/css.md
+++ b/wordpress-coding-standards/css.md
@@ -211,9 +211,11 @@ Incorrect:
 	margin: 0px 0px 20px 0px;
 }
 
-.class {
-	font-family: Times New Roman, serif; /* Quote font names when required */
-	font-weight: bold; /* Avoid named font weights */
+.class { /* Quote font names when required */
+	font-family: Times New Roman, serif;
+	/* Avoid named font weights */
+	font-weight: bold;
+	/* Avoid adding a unit for line height */
 	line-height: 1.4em;
 }
 
--- a/wordpress-coding-standards/css.md
+++ b/wordpress-coding-standards/css.md
@@ -55,24 +55,24 @@ input[type="text"] {
 Incorrect:
 
 [css]
-#commentForm { /* Avoid camelcase. */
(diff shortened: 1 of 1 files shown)


ntwb:
Thanks, I'll have to take a closer look at why the `</em>` is showing there, I suspect the tags are not balanced elsewhere in the file, I'll try to get to that this weekend

secan:
> Thanks, I'll have to take a closer look at why the `</em>` is showing there, I suspect the tags are not balanced elsewhere in the file, I'll try to get to that this weekend

I think it might have something to do with the `*` sign used in the comments (`/*` and `*/`); it appears that the closing asterisk in the first comment is somehow converted into an opening `<em>` tag and the opening one in the following comment becomes the closing `</em>` tag. It is almost like the asterisk is randomly treated as some sort of markdown.

I noticed the same issue appears in the same page, in the section "commenting" (lines 271-276 of _css.md_):
wp_css-coding-standards_commenting

Finally, a similar issue is present in the section "media queries" (lines 241-246 of _css.md_), where GitHub opinionatedly add some `<pre>` and `<code>` tags:
wp_css-coding-standards_media-queries

secan:
I added committed a new version of "css.md" where I escaped all asterisks in the code snippets:
`/*` becomes `/\*`
`*/` becomes`\*/`
this should fix the issue with the `<em>` tags.

I also replaced the pre-formatted text in the "Media Queries" code sample with non formatted text, which should solve the issue of the `<pre>` and `<code>` tags appearing in the sample

ntwb:
Thanks @secan, sorry for the delay, and yes you are most likely correct, as the `*` is used for `<em>` in markdown, which the pages are rendered as markdown and parsed through the Jetpack Markdown Parser this is highly likely the cause, will merge this and see how it looks 💯

ntwb:
This resulted in:

image

I created #41 to use HTML Entities and will see what happens 🤞🏼

PR #41, opened 2020-08-07
Try `&042;` HTML Entity for `*` CSS comments

ntwb:
Follow up to #40 to try using HTML Entity `&042;` for `*` in CSS comments


danfarrow:
This doesn't appear to have worked as expected in the published coding standards: 
https://developer.wordpress.org/coding-standards/wordpress-coding-standards/css/#selectors

PR #42, opened 2020-08-13
Add target blank and rel nofollow on external links

samiahmedsiddiqui:
1. Add target="_blank" on external links and codex links
2. Add rel="nofollow" on external links


ntwb:
Thanks, but a future PR will be removing all HTML code and switch to markdown format for all hyperlinks, as such target blank and no-follow won’t carry over to that

PR #48, opened 2020-10-17
Update markdown_source

hareesh-pillai:
Updated the name of the repository from `docs` to `wpcs-docs`

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -3,76 +3,76 @@
 		"title": "Coding Standards",
 		"slug": "index",
 		"parent": null,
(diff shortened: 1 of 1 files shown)


jrfnl (approved):
LGTM.

Just checking - have you done a search of the codebase to check if the old URL is still used in other files ? If so, it would be great to fix them all in one go.

hareesh-pillai:
Thanks for the review @jrfnl. I found that the old URL is used in another file as well. I've added PR #49 to update that one too.

dingo-d (approved):

PR #55, opened 2021-05-27
Adjust a tab level to fix div & PHP open/close

jonjennings:
We DO indent inside divs and we DON'T indent inside PHP open/close tags - correct?
In which case the indentation here seems incorrect - removing a level of tab from the div fixes both issues

### DIFF
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -182,14 +182,14 @@ Correct (Multiline):
 ```php
 function foo() {
     ?>
-        <div>
+    <div>
         <?php
         echo bar(
             $baz,
             $bat
         );
         ?>
-        </div>
+    </div>
     <?php
 }
 ```


GaryJones (approved):
There's no violation flagged here for `WordPress-Core` or `WordPress-Extra` either before or after the change, but I'm fine with moving the markup left.

jonjennings:
Thanks Gary. Yes I'm routinely checking against WordPress-Core and think it's sometimes a little prone to skipping over indentation issues... in fact I specifically noticed this issue here because I was looking for definitive proof that some code that had already been run through PHPCS was indented incorrectly! I know changing the documentation to match your argument is not a classy move...

jrfnl:
Just FYI: There are no strict indentation rules for "inline HTML", i.e. code outside of the PHP tags.

The reason for that is that the code author may want the resulting HTML to have a logical indentation in the final HTML source code and as this source code can be build up from code snippets spread out across different files, there is just no way for WPCS to keep track of what the "correct" indentation should be.

Hope that explains it a little.

PR #60, opened 2021-09-22
Fix broken link

bazza:
The link to Joseph's blog post no longer works I guess because he changed his permalink structure at some point.  This fixes it.


jrfnl (approved):
Verified. Thanks for fixing this @bazza !

PR #69, opened 2022-05-02
Update accessibility.md

ntwb:
Follow up to #66 to update markdown format for https://developer.wordpress.org/coding-standards/wordpress-coding-standards/accessibility/


jrfnl (approved):

jrfnl:
@ntwb Okay, so I'd left them open for a second reviewer just in case I missed something ;-) Not that there could be much to miss...

Also - could it be that your default "Merge" setting is set to "Rebase & merge" at the moment ? GitHub has this nasty habit of remembering the last "merge type" used (and "normal merge" is the default in this repo IRC ?)

PR #73, opened 2022-05-02
JS Docs: convert from HTML to markdown

jrfnl:
Includes:
* Removing stray unicode artifacts.
* Improving formatting consistency in a couple of places.
    I.e. having "Good"/"Bad" both wrapped in emphasis, having the `:` in lists not included in the bold tags, refering to the correct tags in some examples - most for `@deprecated`.
* Being more liberal with backticks around documentation `@tags`, like in the tables.
* Improving accessibility by not using "here" as link text.
* Removing trailing spaces.

**Notes**:
* The formatting around the `[tip]` should be double-checked as I'm not sure if the shortcode doesn't break now the contents is reformatted.
* There are a few "inline headers" (using `**` not `##..`) in the "Formatting Guidelines" section, about which I wonder if those shouldn't be changed to proper headers.
    See: https://developer.wordpress.org/coding-standards/inline-documentation-standards/javascript/#formatting-guidelines

Also note that while I haven't really "read" the document, only scanned it, I did get the impression that there was quite a lot of redundancy/duplication in the text. Might be something to have a look at/iterate on in a follow-up PR.

### DIFF
--- a/inline-documentation-standards/javascript.md
+++ b/inline-documentation-standards/javascript.md
@@ -1,63 +1,76 @@
 # JavaScript Documentation Standards
 
-WordPress follows the <a href="http://jsdoc.app/" target="_blank" rel="noopener">JSDoc 3 standard</a> for inline JavaScript documentation.
-<h2>What Should Be Documented</h2>
+WordPress follows the [JSDoc 3 standard](http://jsdoc.app/) for inline JavaScript documentation.
+
+## What Should Be Documented
+
 JavaScript documentation in WordPress takes the form of either formatted blocks of documentation or inline comments.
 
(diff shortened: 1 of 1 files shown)


ntwb (approved):
Awesome, thanks @jrfnl 💯

ntwb:
FYI:

> The formatting around the [tip] should be double-checked as I'm not sure if the shortcode doesn't break now the contents is reformatted.

Screenshot of the current `[tip]`, let's see how it looks after the changes are synced:
image

jrfnl:
Just checked - text has synced and `[tip]` looks fine.

PR #74, opened 2022-05-02
PHP Docs: convert from HTML to markdown

jrfnl:
Includes:
* Removing stray unicode artifacts.
* Improving formatting consistency in a couple of places.
    I.e. having the `:` in lists not included in the bold tags, having phrase examples in Good/Bad always in emphasis tags and more.
* Replacing `<u>` tags with `_` (emphasis) tags as markdown does not support underlined text.
* Indented subsequent lines in multi-line ordered list items as the display was not respecting the ordered list numbering now in the ["Description" section](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#description-formerly-long-description).
* Minor punctuation fixes.
* Fixed references to _phpDocumentor_ using the wrong capitalization.
* Made the prohibition against the use of `@return void` clearer by having the complete phrase in backticks instead of just the `@return` tag.
* Removing trailing spaces.

**Notes**:
* The formatting around the `[tip]` and `[info]` shortcodes should be double-checked once the PR is merged & the page is synced.
* In a few places, the text refers to the use of HTML tags. Those tags were written using `&lt;` and `&gt;` HTML encoded text. I've replaced those with `<` and `>` characters, but wrapped the text in backticks so it should be treated as code. Should be double-checked once the PR has been merged & synced that this is displayed correctly in the [Summary](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#summary-formerly-short-description) section.
    This applies to line 78 and 81.
* The link to the source code on line 230 needs verification - it's not completely clear to me to which line in the source this is supposed to refer.
* The formatting of the tag table at the end will have changed as markdown does not allow for multi-line text strings in a table cells. It _should_ wrap automatically though, so :fingers_crossed: the readability should be okay.

I've also made some notes for additional textual changes which IMO should be made, but will pull those in a separate PR once this one has been merged.

### DIFF
--- a/inline-documentation-standards/php.md
+++ b/inline-documentation-standards/php.md
@@ -1,161 +1,162 @@
 # PHP Documentation Standards
 
-WordPress uses a customized documentation schema that draws inspiration from PHPDoc, an evolving standard for providing documentation to PHP code, which is maintained by <a href="http://phpdoc.org/">phpDocumentor</a>.
+WordPress uses a customized documentation schema that draws inspiration from PHPDoc, an evolving standard for providing documentation to PHP code, which is maintained by [phpDocumentor](http://phpdoc.org/).
 
-In some special cases – such as WordPress' implementation of hash notations – standards are derived from the <a href="https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md">draft PSR-5 recommendations</a>. This does not mean we are attempting to be "PSR-5 compliant" at this time, it simply means that we've adopted PSR-5 recommendations <em>in part</em>.
+In some special cases - such as WordPress' implementation of hash notations - standards are derived from the [draft PSR-5 recommendations](https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md). This does not mean we are attempting to be "PSR-5 compliant" at this time, it simply means that we've adopted PSR-5 recommendations _in part_.
 
-<h2>What Should Be Documented</h2>
+## What Should Be Documented
+
 PHP documentation in WordPress mostly takes the form of either formatted blocks of documentation or inline comments.
 
 The following is a list of what should be documented in WordPress files:
-<ul>
- 	<li>Functions and class methods</li>
- 	<li>Classes</li>
- 	<li>Class members (including properties and constants)</li>
- 	<li>Requires and includes</li>
- 	<li>Hooks (actions and filters)</li>
- 	<li>Inline comments</li>
- 	<li>File headers</li>
- 	<li>Constants</li>
-</ul>
-<h3>Documenting Tips</h3> 
-<h4>Language</h4>
+
+- Functions and class methods
+- Classes
+- Class members (including properties and constants)
+- Requires and includes
+- Hooks (actions and filters)
+- Inline comments
+- File headers
+- Constants
+
+### Documenting Tips
+
+#### Language
 
 Summaries (formerly Short Descriptions) should be clear, simple, and brief. Avoid describing "why" an element exists, rather, focus on documenting "what" and "when" it does something.
 
-A function, hook, class, or method is a <em>third-person singular</em> element, meaning that <em>third-person singular verbs</em> should be used to describe what each does.
+A function, hook, class, or method is a _third-person singular_ element, meaning that _third-person singular verbs_ should be used to describe what each does.
 
-[tip]Need help remembering how to conjugate for third-person singular verbs? Imagine prefixing the function, hook, class, or method summary with "It":
-<ul>
- 	<li><em>Good:</em> "(It) Does something."</li>
- 	<li><em>Bad:</em> "(It) Do something."</li>
-</ul>
+[tip]
+Need help remembering how to conjugate for third-person singular verbs? Imagine prefixing the function, hook, class, or method summary with "It":
+
+- _Good_: "(It) Does something."
+- _Bad_: "(It) Do something."
+
 [/tip]
 
 Summary examples:
-<ul>
- 	<li><strong>Functions:</strong> <u>What</u> does the function do?
-<ul>
- 	<li>Good: <em>Displays the last modified date for a post.</em></li>
- 	<li>Bad: <em>Display the date on which the post was last modified.</em></li>
-</ul>
-</li>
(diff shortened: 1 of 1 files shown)


jrfnl:
Original Summary section for double-check later:
image

inline-documentation-standards/php.md
```diff
@@ -213,25 +216,26 @@ Functions and class methods should be formatted as follows:
...
+Array values should be documented using WordPress' flavor of hash notation style similar to how [Hooks](http://make.wordpress.org/core/handbook/inline-documentation-standards/php-documentation-standards/#4-hooks-actions-and-filters) can be documented, each array value beginning with the `@type` tag, and taking the form of:
 
 ```php
 *     @type type $key Description. Default 'value'. Accepts 'value', 'value'.
 *                     (aligned with Description, if wraps to a new line)
 ```
 
-An example of an "originating" function and re-use of an argument array is <a href="https://core.trac.wordpress.org/browser/branches/4.0/src/wp-includes/http.php#L115"><code>wp_remote_request|post|get|head()</code></a>.
+An example of an "originating" function and re-use of an argument array is [`wp_remote_request|post|get|head()`](https://core.trac.wordpress.org/browser/branches/4.0/src/wp-includes/http.php#L115).
```
jrfnl, line 230: This is the link of which I'm not so sure it is pointing to the correct place.

jrfnl:
Incorrect numbering in the Description section (as it was):

image

inline-documentation-standards/php.md
dingo-d: Maybe `https://core.trac.wordpress.org/browser/trunk/src/wp-includes/http.php#L114` would be better? But still, if anything changes in this file that would cause the line change the docs should be updated.

Maybe we could just link to the `http.php` file and say, check the doc block for the `wp_remote_request` function?

jrfnl:
Anything I can do to move this PR forward ?

inline-documentation-standards/php.md
GaryJones: Why not just link to a single function's page?

```suggestion
An example of an "originating" function and re-use of an argument array is [`wp_remote_request()`](https://developer.wordpress.org/reference/functions/wp_remote_request/).
```

GaryJones (approved):
Conversion looks fine. Left one comment about the content.

Why is this conversion being done in the first place?

inline-documentation-standards/php.md
jrfnl: The way I read it - the other functions are re-using the information in the docblock for `wp_remote_request()` by pointing to that one using a `@see`. So just pointing to the one function would not illustrate that well enough.

jrfnl:
> Why is this conversion being done in the first place?

Good question and I don't have a great answer.

From my perspective:
* Converting the HTML to markdown "should" probably have been done when the handbook editing got moved to this repo, but most docs never did get converted until now.
* As we're going to be doing some heavy editing of a few of these docs for WPCS 3.0.0, we may as well do it now and do it in one go for all docs which haven't been converted yet.
* Using markdown instead of HTML makes for cleaner docs to maintain.
* Once everything is markdown, we can maintain consistency with `markdownlint`.

@dingo-d and @ntwb may have some additional thoughts on this.

dingo-d:
Agree with Juliette on this one, since we already started moving everything to markdown I don't see an issue with moving these docs as well.

inline-documentation-standards/php.md
dingo-d: @jrfnl we can just change it to point to the 6.0 branch, and check for it periodically every 5-6 'minor' versions to see if the line number changed.

ntwb:
> Why is this conversion being done in the first place?

All the sub-sites of https://developer.wordpress.org/ are generated from markdown and imported via [this dotorg plugin](https://github.com/WordPress/wordpress.org/tree/trunk/wordpress.org/public_html/wp-content/plugins/wporg-markdown) which uses the [Jetpack markdown parser](https://github.com/Automattic/jetpack/tree/master/projects/plugins/jetpack/_inc/lib/markdown), so importing via native markdown helps with avoid HTML parsing issues we've had in the past and also makes readability of the files here in this repo a little easier to read and update when contributing changes.

inline-documentation-standards/php.md
ntwb: I think as this is for PHP Docs, we want to highlight how the array parameters are displayed in the code reference, not so much the actual PHPDcs in the source code.

So, linking to https://developer.wordpress.org/reference/functions/wp_remote_request/#return should do the trick, I think and no versioning of line numbers required, this anchor will always be the same

image

jrfnl: I've had another look at this, but I think linking to the develop docs is the wrong course of action. Please read the whole section to see what the link is trying to illustrate.

I believe this is the crucial sentence:

> Parameters that are an array of arguments should be documented in the "originating" function only, and cross-referenced via an `@see` tag in corresponding DocBlocks.

Pointing to the dev docs would do nothing to illustrate how someone updating the docblocks would need to apply this.

dingo-d: So you'd agree that linking to trunk, or even the latest release, would be a better thing to do?

jrfnl: Linking to `trunk` would not be stable as line numbers change, linking to the `6.0` release would definitely be better, but what line number should it link to ?

dingo-d: So the original suggestion linked to the `wp_remote_request`, but there is no mention of what is the arguments array actually contains unless you go and dig deeper.

If we want to link to what the contents of the arguments array are, then the correct link would be:

https://core.trac.wordpress.org/browser/tags/6.0/src/wp-includes/class-wp-http.php#L105

jrfnl: I've updated the link now to: https://core.trac.wordpress.org/browser/tags/6.0/src/wp-includes/http.php/#L114 (same file)

jrfnl:
Would love to (finally) see this one merged. Could it please get some review love ?

dingo-d (approved):

jrfnl:
Just verified the screenshotted parts of the page now the update has been merged and IM all looks good. 🎉

PR #70, opened 2022-05-02
Update css.md

ntwb:
Follow up to #66 to update markdown format for https://developer.wordpress.org/coding-standards/wordpress-coding-standards/css/

PR #76, opened 2022-05-18
Added online version of autoprefixer

rahulkumarsingh73690:
Added online version of Autoprefixer. So users can instantly use it without installing it.

### DIFF
--- a/wordpress-coding-standards/css.md
+++ b/wordpress-coding-standards/css.md
@@ -150,7 +150,7 @@ Example:
 <h3>Vendor Prefixes</h3>
 Updated on 2/13/2014, after <a href="https://core.trac.wordpress.org/changeset/27174">[27174]</a>:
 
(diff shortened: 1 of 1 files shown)


jrfnl:
Without having an opinion on this change: I noticed you changed the link from markdown to HTML. Just FYI - all the docs here are being converted to be markdown only. There is a PR currently open to do so for this document as well: #70

rahulkumarsingh73690:
@jrfnl 

I saw all links in href format that’s why I changed this link. 

Sorry, I changed href to md link. If you allow I will convert entire page from html to markdown

jrfnl:
> If you allow I will convert entire page from html to markdown

There is already a PR open which does just that - see PR #70.

rahulkumarsingh73690:
@jrfnl 

Can you please review this PR

dingo-d (commented):
Isn't this more official link: https://autoprefixer.github.io/ ?

rahulkumarsingh73690:
@dingo-d

None of these are official but I recommend this one (https://goonlinetools.com/autoprefixer/) because it highlights css code error and both works on latest autoprefixer and postcss.

ntwb:
Thanks, but no thanks, disclosing you're relationship that you're the developer of that site @rahulkumarsingh73690 would have been beneficial upfront

I'm closing this as if developers would like to learn more about Autoprefixer they can look at the official repo and discover additional resources from the extensive readme Autoprefixer includes

PR #65, opened 2022-04-17
Content update changes in PHP Coding Standards

dingo-d:
This is the first part of the updates to the WPCS PHP documentation. In this PR I've gone through the text and replaced all the HTML with markdown syntax.

Some links have been updated - some HTTP to HTTPS replacements; data validation now points to the plugins handbook (as a part of the move from codex to developers documentation); added title refs to trac tickets.

I've looked at the JS coding standards documentation and markdown is used there, so don't see why it shouldn't be used here as well.

The second part of the updates will see what rules can be added from the [make post](https://make.wordpress.org/core/2020/03/20/updating-the-coding-standards-for-modern-php/) in 2020. There are some ongoing discussions that are not resolved, but I'll add only the one that nobody had any issues with within the comments.

@jrfnl and I agreed that we should add this to the handbook so that the WPCS 3.0.0 release can be made in line with the new additions. 

But I'll open a separate PR for that 🙂 

Fixes: #19

PR #92, opened 2022-06-17
Fix the code examples

dingo-d:
This PR contains the suggestions and fixes for the code examples based on comments from #65 PR.

Mostly fixed the escaping issues, some comments missing full stops at the end, and the like.

### DIFF
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -454,7 +454,7 @@ function eat( $what, $args ) {
 eat ( 'noodles', array( 'speed' => 'moderate' ) );
 ```
 
-Be careful when using this pattern, as it can lead to "Undefined array index" notices in your code if proper checks are not set in place. Use this pattern only where it makes sense (i.e. multiple possible arguments), not just for the sake of it.
+Be careful when using this pattern, as it can lead to "Undefined array index" notices if input isn't validated before use. Use this pattern only where it makes sense (i.e. multiple possible arguments), not just for the sake of it.
 
 ## Interpolation for Naming Dynamic Hooks
 
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -521,7 +521,7 @@ Unless absolutely necessary, loose comparisons should not be used, as their beha
 Correct:
 
 ```php
-if ( 0 === strpos( 'WordPress', 'foo' ) ) {
+if ( 0 === strpos( $text, 'WordPress' ) ) {
 	echo esc_html__( 'Yay WordPress!' );
 }
 ```
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -16,7 +16,7 @@ Use single and double quotes when appropriate. If you're not evaluating anything
 
 ```php
 echo '<a href="/static/link" class="button button-primary">Link name</a>';
-echo "<a href='$escaped_link'>text with a ' single quote</a>";
+echo "<a href='{$escaped_link}'>text with a ' single quote</a>";
 ```
 
 Text that goes into HTML or XML attributes should be escaped so that single or double quotes do not end the attribute value and invalidate the HTML, causing a security issue. See [Data Validation](https://developer.wordpress.org/plugins/security/data-validation/) in the Plugin Handbook for further details.
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -559,7 +559,7 @@ In a `switch` statement, it's okay to have multiple empty cases fall through to 
 switch ( $foo ) {
 	case 'bar':	      // Correct, an empty case can fall through without comment.
 	case 'baz':
-		echo esc_html( $foo );    // Incorrect, a case with a block must break, return, or have a comment.
+		echo esc_html( $foo ); // Incorrect, a case with a block must break, return, or have a comment.
 	case 'cat':
 		echo 'mouse';
 		break;        // Correct, a case with a break does not require a comment.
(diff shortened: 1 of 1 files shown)


wordpress-coding-standards/php.md
```diff
@@ -192,16 +192,18 @@ Correct (Multiline):
...
+		<?php
+		echo esc_html(
+			bar(
+				$baz,
+				$bat
+			)
+		);
+		?>
+	</div>
+	<?php
```
jrfnl, line 206: This may need a check that the tabs in code examples in markdown are handled correctly when the page gets converted to the handbook.

In my experience spaces are the way to go with markdown, but this is kind of a special set up, so I'm not sure.

dingo-d:
@ntwb should I replace all the tabs in the code examples with spaces as well? The code examples on the handbook, when copied will always paste as spaces in the editors, and I'm not sure what's the preferred option when writing markdown code (using tabs or spaces).

wordpress-coding-standards/php.md
```diff
@@ -446,13 +448,14 @@ eat( 'dogfood', 'quickly' );
...
 
 ```php
-// Even Better
 function eat( $what, $args ) {
 ...
 }
 eat ( 'noodles', array( 'speed' => 'moderate' ) );
 ```
 
+Be careful when using this pattern, as it can lead to "Undefined array index" notices in your code if proper checks are not set in place. Use this pattern only where it makes sense (i.e. multiple possible arguments), not just for the sake of it.
```
jrfnl, line 457:
```suggestion
Be careful when using this pattern, as it can lead to "Undefined array index" notices if input isn't validated before use. Use this pattern only where it makes sense (i.e. multiple possible arguments), not just for the sake of it.
```

wordpress-coding-standards/php.md
```diff
@@ -519,15 +522,15 @@ Correct:
 
 ```php
 if ( 0 === strpos( 'WordPress', 'foo' ) ) {
```
jrfnl, line 524:
```suggestion
if ( 0 === strpos( $text, 'WordPress' ) ) {
```

While I know it is only a code example, that condition could never ever match, so was a bit nonsensical. This makes for a more realistic example.

(here and in the below "incorrect" example)

wordpress-coding-standards/php.md
```diff
@@ -556,7 +559,7 @@ In a `switch` statement, it's okay to have multiple empty cases fall through to
 switch ( $foo ) {
 	case 'bar':	      // Correct, an empty case can fall through without comment.
 	case 'baz':
-		echo $foo;    // Incorrect, a case with a block must break, return, or have a comment.
+		echo esc_html( $foo );    // Incorrect, a case with a block must break, return, or have a comment.
```
jrfnl, line 562:
```suggestion
		echo esc_html( $foo ); // Incorrect, a case with a block must break, return, or have a comment.
```

(or maybe the comments should be re-aligned ?)

jrfnl (commented):
Thanks @dingo-d for this follow-up!

Just left a few small remarks.

Note: I haven't gone through the full text document again as I trust you have (and we all did in the previous thread).

dingo-d:
@GaryJones @ntwb can you give a quick look at this so that I can move on with #93?

Thanks!

wordpress-coding-standards/php.md
```diff
@@ -518,16 +521,16 @@ Unless absolutely necessary, loose comparisons should not be used, as their beha
 Correct:
 
 ```php
-if ( 0 === strpos( 'WordPress', 'foo' ) ) {
-	echo __( 'Yay WordPress!' );
+if ( 0 === strpos( $text, 'WordPress' ) ) {
+	echo esc_html__( 'Yay WordPress!' );
```
GaryJones, line 525: Should one of these examples have a textdomain?

GaryJones (approved):
Approved, albeit with one query.

wordpress-coding-standards/php.md
dingo-d: I'd leave it without because it's mostly intended to be used for WP core.

Do you agree with that @GaryJones ?

GaryJones: > The purpose of the WordPress Coding Standards is to create a baseline for collaboration and review within various aspects of the WordPress open source project and community, from core code to themes to plugins.
>
> The WordPress community developed the standards contained in this section of the handbook, and those standards are part of the best practices that developers and core contributors are recommended to follow.
> — https://developer.wordpress.org/coding-standards/wordpress-coding-standards/ 

I bet there would be far more non-Core developers (i.e. plugin and theme developers) looking at these docs, than WP Core developers or Core contributors, so if we're going to emphasise best practices and expectations, and we have examples like [this](https://github.com/WordPress/wpcs-docs/pull/92/files#diff-b06b02910dcd169873e32c65fc4a0528d9b7ba05e553e2718f57b88d2bd5ab88R169) which already includes textdomains, then it would make sense to be consistent?

GaryJones: We're also using `esc_html__()` here, yet Core would use `__()` as they don't consider bad/suspicious translations to be a risk, so it's clearly tailored towards non-Core developers; not including the textdomain makes it an example that isn't accurate for either group of developers.

dingo-d: I can add a textdomain, not a big problem 🙂 

Would this be better:

```php
	echo esc_html__( 'Yay WordPress!', 'textdomain' );
```

Or should we do something more specific for the textdomain name?

dingo-d:
@jrfnl when you'll have the time can you check if the addition is ok?

wordpress-coding-standards/php.md
jrfnl: No need for changes, but I just noticed this uses a Yoda condition in a place where it isn't needed as neither side of the comparison is a variable, so this condition could just be written like `if ( strpos( $text, 'WordPress' ) === 0 ) {`

jrfnl (approved):
Aside from the open question about tabs vs spaces in the code samples, all looks good to me.

PR #97, opened 2022-08-12
Add namespace declaration section

dingo-d:
This is the first PR that aims to add the additions of 'modern' coding standards in the WP PHP Coding Standards handbook based on the [make post](https://make.wordpress.org/core/2020/03/20/updating-the-coding-standards-for-modern-php/) created by Juliette Reinders Folmer over two years ago.

The aim is to add some 15ish new additions from the post (maybe less because some proposals do need additional discussion) to the handbook so that we have guides in place for anybody who uses modern code additions like namespaces and traits in their code, but wishes to align it to WordPress Coding Standards rules.

This will also enable us to add the rules to the wpcs sniffs so that these checks are automated across the WordPress Core as well. 

I tried to distill the make post so that the handbook contains examples and readable information about how things should be written, without being too technical like in the make post.

Comments are more than welcome.

### DIFF
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -366,6 +366,47 @@ $a = foo(
 );
 ```
 
+## Declare Statements, Namespace, and Import Statements
+
+### Namespace declarations
+
+Each part of a namespace name should consist of capitalized words separated by underscores.
+
+Namespace declarations should have exactly one blank line before the declaration and at least one blank line after.
+
+```php
+namespace Prefix\Admin\Domain_URL\Sub_Domain\Event; // Correct.
+```
+
+There should be only one namespace declaration per file, and it should be at the top of the file. Namespace declarations using curly brace syntax are not allowed. Explicit global namespace declaration (namespace declaration without name) are also not allowed.
+
+```php
+// Incorrect: namespace declaration using curly brace syntax.
+namespace Foo {
+    // Code.
+}
+
+// Incorrect: namespace declaration for the global namespace.
+namespace {
+    // Code.
+}
+```
+
+_There is currently no timeline for introducing namespaces to WordPress Core._ 
+
+The use of namespaces in plugins and themes is strongly encouraged. It is a great way to prefix a lot of your code to prevent naming conflicts with other plugins, themes and/or WordPress Core.
+
+Please do make sure you use a unique and long enough namespace prefix to actually prevent conflicts. Generally speaking, using a namespace prefix along the lines of `Vendor\Project_Name` is a good idea.
+
+[warning]
+The `wp` and `WordPress` namespace prefixes are reserved for WordPress itself.
+[/warning]
+
+[info]
+Namespacing has no effect on variables, constants declared with `define()` or non-PHP native constructs, like the hook names as used in WordPress.
+Those still need to be prefixed individually.
+[/info]
+
 ## Object-Oriented Programming
 
 ### Only One Object Structure (Class/Interface/Trait) per File


wordpress-coding-standards/php.md
```diff
@@ -135,6 +135,62 @@ do_action( "{$new_status}_{$post->post_type}", $post->ID, $post );
...
+namespace Foo {
+    // Code.
+}
+
+/**
+ * Global namespace.
+ */
+namespace {
+    // Code.
+}
```
jrfnl, line 178:
```suggestion
// Incorrect: namespace declaration using curly brace syntax.
namespace Foo {
    // Code.
}

// Incorrect: namespace declaration for the global namespace.
namespace {
    // Code.
}
```

* The `<?php` in the code samples on Make were needed for the syntax highlighting to work, but aren't typically used in the handbook (unless they are part of the code sample).
* The example with the comments can IMO be left out of the handbook as it is a parse error in PHP 8.0+ anyway (and let's not give people ideas).

wordpress-coding-standards/php.md
```diff
@@ -135,6 +135,62 @@ do_action( "{$new_status}_{$post->post_type}", $post->ID, $post );
...
+
+_There is currently no timeline for introducing namespaces to WordPress Core._ 
+
+The use of namespaces in plugins and themes is strongly encouraged. It is a great way to prefix a lot of your code to prevent naming conflicts with other plugins, themes and/or WordPress Core.
+
+Please do make sure you use a unique and long enough namespace prefix to actually prevent conflicts. Generally speaking, using a namespace prefix along the lines of Vendor\Packagename is a good idea.
+
+[warning]
+Using `wp` or `WordPress` as a prefix is not allowed.
+[/warning]
```
jrfnl, line 189: I don't think I've seen the `[warning]` shortcode used like this before ? Does this work ? How would this display ?

(in other places I've seen the more "markdowny"-syntax: `> Warning: ... `)


Just thinking: maybe this bit should actually be moved up to the first paragraph which talks about naming conventions ?

Something like: _The `wp` and `WordPress` namespace prefixes are reserved for WordPress itself._

wordpress-coding-standards/php.md
```diff
@@ -135,6 +135,62 @@ do_action( "{$new_status}_{$post->post_type}", $post->ID, $post );
...
+
+Namespacing has no effect on variables declared outside of functions/classes, constants declared with `define()` or non-PHP native constructs, like the hook names as used in WordPress.
+Those still need to be prefixed individually.
```
jrfnl, line 192: I don't think this belongs in this section as it:
1. Is not actually about coding conventions for namespaces.
2. Is only applicable for plugins/themes.

jrfnl (changes requested):
@dingo-d Thanks for this PR! Great to see the next steps being taken on turning the proposal into reality!

Suggestion: move the section down to just before the "Object Oriented" section. It feels a bit odd having this before the (very generic) "Whitespace" section.

wordpress-coding-standards/php.md
dingo-d: The inline JS coding standards are using some callout shortcodes here: https://developer.wordpress.org/coding-standards/inline-documentation-standards/javascript/ (they are using `[tip]` ones).

I thought about adding the tips at the top, but it made more sense to me to leave them after we've introduced the namespaces. First the definition, then the do's/don'ts

wordpress-coding-standards/php.md
dingo-d: I agree with the part that it's not a coding convention, but more of useful knowledge, and since we are recommending the usage of the standards to all the WP community (outside of core).

It's similar to what we added in the Recommendation section about named arguments in my opinion. Plus extra knowledge cannot hurt, and I'm positive that a lot of WP devs are not aware of that fact 😄

jrfnl: In that case, what about putting this in an `[info]` block ?

wordpress-coding-standards/php.md
jrfnl: Fair enough - still may be good to adjust the phrasing to my suggestion ?

wordpress-coding-standards/php.md
dingo-d: Love that idea!

wordpress-coding-standards/php.md
```diff
@@ -366,6 +366,43 @@ $a = foo(
...
+namespace {
+    // Code.
+}
+```
+
+_There is currently no timeline for introducing namespaces to WordPress Core._ 
+
+The use of namespaces in plugins and themes is strongly encouraged. It is a great way to prefix a lot of your code to prevent naming conflicts with other plugins, themes and/or WordPress Core.
+
+Please do make sure you use a unique and long enough namespace prefix to actually prevent conflicts. Generally speaking, using a namespace prefix along the lines of Vendor\Packagename is a good idea.
```
GaryJones, line 397:
```suggestion
Please do make sure you use a unique and long enough namespace prefix to actually prevent conflicts. Generally speaking, using a namespace prefix along the lines of Vendor\ProjectName is a good idea.
```

- The "package name" (as in what Composer might use) would be the [combination](https://getcomposer.org/doc/01-basic-usage.md#package-names) of the vendor + project name, so let's avoid using the word "package" here.
- Namespaces can include capitals mid-word, so let's show that as a valid example, rather than the title-case example that was currently used.

jrfnl: Shouldn't it in that case be `Vendor\Project_Name` (if we're talking compliance with the WP naming conventions for namespaces which we're adding in this PR) ?

wordpress-coding-standards/php.md
```diff
@@ -366,6 +366,45 @@ $a = foo(
...
+The use of namespaces in plugins and themes is strongly encouraged. It is a great way to prefix a lot of your code to prevent naming conflicts with other plugins, themes and/or WordPress Core.
+
+Please do make sure you use a unique and long enough namespace prefix to actually prevent conflicts. Generally speaking, using a namespace prefix along the lines of `Vendor\Project_Name` is a good idea.
+
+[warning]
+The `wp` and `WordPress` namespace prefixes are reserved for WordPress itself.
+[/warning]
+
+[info]
+Namespacing has no effect on variables declared outside of functions/classes, constants declared with `define()` or non-PHP native constructs, like the hook names as used in WordPress.
```
jrfnl, line 404:
```suggestion
Namespacing has no effect on variables, constants declared with `define()` or non-PHP native constructs, like the hook names as used in WordPress.
```

jrfnl (approved):
💯

jrfnl:
Just double-checking it against the Make post now - only thing I can see that's missing is the "one blank line before and after a namespace declaration"....

dingo-d:
I've added the blank line note, can you check the wording?

jrfnl (approved):

dingo-d:
@GaryJones when you'll find time, please review this, and if ok, merge it so that I can rebase the rest of the PRs.

It's going to be easier to review the others one by one after each previous has been reviewed and merged.

wordpress-coding-standards/php.md
```diff
@@ -366,6 +366,47 @@ $a = foo(
 );
 ```
 
+## Declare Statements, Namespace, and Import Statements
+
+### Namespace declarations
+
+Each part of a namespace name should consist of capitalized words separated by underscores.
```
GaryJones, line 373: > separated by underscores

Does this originate from the WP standard being to name classes this way as well?

I agree that camel case namespaces and underscored class names would look very wrong, but this namespace puts us at direct odds with PSR-0/4.

GaryJones (approved):
✅  Approved with one observation.

wordpress-coding-standards/php.md
jrfnl: From the Make post:
> Each “leaf” of a namespace name should be in Camel_Caps, i.e. each word should start with a capital and words should be separated by an underscore.
> Consecutive caps, like for acronyms, will be allowed. This is in line with other WP naming conventions.

The class file naming scheme already makes PSR-0/4 impossible anyway, so when I created the make post, I favoured consistency with existing WP rules for class names (principle of least surprise).

The one place where I strayed from that (spacing in `declare` statements), straight away got the (expected) remarks about introducing inconsistency with pre-existing rules.

PR #99, opened 2022-08-12
PHP Docs: various small textual updates

jrfnl:
* Remove reference to PSR5 as the current version of PSR5 is pretty different from the PSR5 draft which was used at the time this was written.
* Remove outdated naming references.
* Clarify when `@access` should be used (now often used incorrectly in Core as it isn't specified what is meant by "in limited circumstances").
* Clarify the correct use of `{@internal ...}.
* Make the list of deprecated tags a little more complete.
* Add link to draft PSR19.


dingo-d (approved):

GaryJones (approved):
✅

PR #98, opened 2022-08-12
Add import use statements section

dingo-d:
The PR adds rules about import `use` statements in the code and is the continuation of the additions of 'modern' PHP code in the WordPress PHP Coding Standards handbook based on the [make post](https://make.wordpress.org/core/2020/03/20/updating-the-coding-standards-for-modern-php/) by Juliette Reinders Folmer.

I'd hold off of a review until the above PR is merged.

PR #100, opened 2022-08-12
Use fully qualified names in inline code

dingo-d:
This PR depends on #98. Once that is merged, this PR should be rebased to the master branch.

The PR adds rules about the fully qualified names in the inline code and is the continuation of the additions of 'modern' PHP code in the WordPress PHP Coding Standards handbook based on the [make post](https://make.wordpress.org/core/2020/03/20/updating-the-coding-standards-for-modern-php/) by Juliette Reinders Folmer.

### DIFF
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -482,6 +482,24 @@ While import `use` statements can already be used in WordPress Core, it is, for 
 Import `use` statements are most useful when combined with namespaces and a class autoloading implementation.
 As neither of these are currently in place for WordPress Core and discussions about this are ongoing, holding off on adding import `use` statements to WordPress Core is the sensible choice for now.
 
+### Using fully qualified names in inline code
+
+When using a fully or partially qualified name, no whitespace or comments are allowed within the name.
+
+```php
+// Correct.
+$foo = new \Vendor\Domain\Foo();
+
+$result = namespace\function_name();
+
+// Incorrect.
+$foo = new \Vendor // Source: external dependency.
+           \Domain
+           \Foo();
+
+$result = namespace \function_name(); // Notice the space between namespace and \function_name().
+```
+
 ## Object-Oriented Programming
 
 ### Only One Object Structure (Class/Interface/Trait) per File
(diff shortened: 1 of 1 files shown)


wordpress-coding-standards/php.md
```diff
@@ -407,6 +407,100 @@ Namespacing has no effect on variables, constants declared with `define()` or no
...
+// Incorrect.
+$foo = new \Vendor // Source: external dependency.
+           \Domain
+           \Foo();
+
+$result = namespace \function_name(); // Notice the space between namespace and \function_name().
+```
+
+>>>>>>> cfb9817... Add fully qualified names in inline code chapter
 ## Object-Oriented Programming
```
jrfnl, line 504: Merge conflict artifact in the commit

jrfnl (commented):
I'm not sure this needs to be added to the manual. As of PHP 8.0, using whitespace or comments within the name is a parse error (remember: the Make post was written well before PHP 8.0 was released).

dingo-d:
> I'm not sure this needs to be added to the manual. As of PHP 8.0, using whitespace or comments within the name is a parse error (remember: the Make post was written well before PHP 8.0 was released).

I can add it as a note below, would be a good addition IMO 👍🏼

jrfnl (commented):
While the note about PHP 8.0 does help, I still don't think this section needs to be in the manual.

Code shouldn't cause parse errors on any of the supported PHP versions, Doesn't matter the reason of the parse error.

That's not something which is a coding standard, that's something which is a basic necessity for the code to be able to run.

dingo-d:
Oh, you meant we should drop this all together? I just added all the rules we had in the spreadsheet 😂 

But your reasoning makes complete sense. I'll close this PR 👍🏼

GaryJones:
Was there any proposal about when to use inline FQCN vs an import?

jrfnl:
There was not. At the time, we decided that as WP isn't using namespaces/import `use` statements, it was not up to WP to set a guideline/standard for when to use FCNs vs import statements.

I know of at least one company in the WP sphere were they strictly enforce always importing classes, but forbid importing global functions/constants and instead enforce using FQNs for those. For namespaced functions/constants, the namespace is imported, and the functions/constants use partially qualified names.

Both Slevomat as well as PHPCSExtra have various sniffs which could enforce one or the other, so userland has a choice here (and the means to enforce their choice).

PR #102, opened 2022-08-12
Type declarations

dingo-d:
This PR depends on #101. Once that is merged, this PR should be rebased to the master branch.

The PR adds rules about the type declaration rules and is the continuation of the additions of 'modern' PHP code in the WordPress PHP Coding Standards handbook based on the [make post](https://make.wordpress.org/core/2020/03/20/updating-the-coding-standards-for-modern-php/) by Juliette Reinders Folmer.

### DIFF
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -385,6 +385,62 @@ $a = foo(
 );
 ```
 
+### Type declarations
+
+Type declarations must have exactly one space before and after the type. The nullability operator (`?`) is regarded as part of the type declaration and there should be no space between this operator and the actual type. Class/interface/enum name based type declarations should use the case of the class/interface/enum name as declared, while the keyword-based type declarations should be lowercased.
+
+Return type declarations should have no space between the closing parenthesis of the function declaration and the colon starting a return type.
+
+These rules apply to all structures allowing for type declarations: functions, closures, enums, catch conditions as well as the PHP 7.4 arrow functions and typed properties.
+
+```php
+// Correct.
+function foo( Class_Name $parameter, callable $callable, int $number_of_things = 0 ) {
+    // Do something.
+}
+
+function bar(
+    Interface_Name&Concrete_Class $param_a,
+    string|int $param_b,
+    callable $param_c = 'default_callable'
+): User|false {
+    // Do something.
+}
+
+// Incorrect.
+function baz(Class_Name $param_a, String$param_b,      CALLABLE     $param_c )   :   ?   iterable   {
+    // Do something.
+}
+```
+
+[info]
+Type declaration usage has the following restrictions based on the minimum required PHP version of an application, whether it is WordPress Core, a plugin or a theme:
+
+- The scalar `bool`, `int`, `float`, and `string` type declarations cannot be used until the minimum PHP version is PHP 7.0 or higher.
+- Return type declarations cannot be used until the minimum PHP version is PHP 7.0 or higher.
+- Nullable type declarations cannot be used until the minimum PHP version is PHP 7.1 or higher.
+- The `iterable` and `void` type declarations cannot be used until the minimum PHP version is PHP 7.1 or higher. The `void` type can only be used as a return type.
+- The `object` type declaration cannot be used until the minimum PHP version is PHP 7.2 or higher.
+- Property type declarations cannot be used until the minimum PHP version is PHP 7.4 or higher.
+- `static` (return type only) cannot be used until the minimum PHP version is PHP 8.0 or higher.
+- `mixed` type cannot be used until the minimum PHP version is PHP 8.0 or higher. Take note that the `mixed` type includes `null`, so cannot be made nullable.
+- Union types cannot be used until the minimum PHP version is PHP 8.0 or higher.
+- Intersection types cannot be used until the minimum PHP version is PHP 8.1 or higher.
+- `never` (return type only) cannot be used until the minimum PHP version is PHP 8.1 or higher.
+- Disjunctive normal form types (combining union types and intersection types) cannot be used until the minimum PHP version is PHP 8.2 or higher.
+[/info]
+
+_Usage in WordPress Core_
+
+[warning]
+Adding type declarations to existing WordPress Core functions should be done with extreme care.
+[/warning]
+
+The function signature of any function (method) which can be overloaded by plugins or themes should not be touched.
+This leaves, for now, only unconditionally declared functions in the global namespace, `private` class methods, and code new to Core, as candidates for adding type declarations.
+
+Note: Using the `array` keyword in type declarations is **strongly discouraged** for now, as most often, it would be better to use `iterable` to allow for more flexibility in the implementation and that keyword is not yet available for use in WordPress Core until the minimum requirements are raised to PHP 7.1.
+
 ## Declare Statements, Namespace, and Import Statements
 
 ### Namespace declarations


wordpress-coding-standards/php.md
```diff
@@ -366,6 +366,263 @@ $a = foo(
...
+    ...
+}
+```
+
+### Type declarations
+
+[alert]
+In WordPress Core, type declaration usage has the following restrictions based on the minimum required PHP version:
+
+- The scalar `bool`, `int`, `float` and `string` type declarations can not be used until the minimum PHP version has been raised to PHP 7.0.
```
GaryJones, line 417:
```suggestion
- The scalar `bool`, `int`, `float`, and `string` type declarations can not be used until the minimum PHP version has been raised to PHP 7.0.
```

wordpress-coding-standards/php.md
```diff
@@ -366,6 +366,263 @@ $a = foo(
...
+- Return type declarations can not be used until the minimum PHP version has been raised to PHP 7.0.
+- Nullable type declarations can not be used until the minimum PHP version has been raised to PHP 7.1.
+- The `iterable`, `void` and `object` type declarations can not be used until the minimum PHP version has been raised to PHP 7.1 and 7.2 respectively.
```
GaryJones, line 420:
```suggestion
- The `iterable`, `void`, and `object` type declarations can not be used until the minimum PHP version has been raised to PHP 7.1 and 7.2 respectively.
```

Alternatively, split these up into separate bullets for PHP 7.1 and PHP 7.2.

wordpress-coding-standards/php.md
```diff
@@ -366,6 +366,263 @@ $a = foo(
...
+Property type declarations can not be used until the minimum PHP version has been raised to PHP 7.4.
```
GaryJones, line 421:
```suggestion
- Property type declarations can not be used until the minimum PHP version has been raised to PHP 7.4.
```

wordpress-coding-standards/php.md
```diff
@@ -366,6 +366,263 @@ $a = foo(
...
+    public $baz = true; // Missing blank line after trait import.
+
+    ...
+}
+```
+
+### Type declarations
+
+[alert]
+In WordPress Core, type declaration usage has the following restrictions based on the minimum required PHP version:
```
GaryJones, line 415: > In WordPress Core,...

Surely this also applies to any plugin or theme as well, not just WP core?

wordpress-coding-standards/php.md
```diff
@@ -366,6 +366,263 @@ $a = foo(
...
+// Correct.
+function foo( Class_Name $param_a, ?string $param_b, callable $param_c ): ?\Class_Name {
+    // Do something.
+}
+
+function bar(
+    Interface_Name $param_a,
+    ?int $param_b,
+    bool $param_c
+): ?\Class_Name {
```
GaryJones, line 465:
```suggestion
): iterable {
```

Just to mix it up a little as the `foo()` function above also returns the exact same type, and it's good to see a non-null, non-FQCN example.

wordpress-coding-standards/php.md
dingo-d: Valid point 👍🏼 I will remove the part about the WP Core.

wordpress-coding-standards/php.md
dingo-d: Splitting these might make more sense, seeing how others are split by versions 👍🏼

dingo-d:
I've added all the types currently available that cannot be used in core (still), but since the wording in the alert now excludes core I've reworded the phrase `until the minimum version is raised` to `until the minimum PHP version is ... or higher`. Because it didn't really make sense (as there is no mention of WordPress, so where the PHP minimum required version comes from wasn't really clear to me in the sentence context).

I'm open to rewording these statements if they read weird in any way.

wordpress-coding-standards/php.md
```diff
@@ -370,6 +370,91 @@ $a = foo(
...
+- Return type declarations cannot be used until the minimum PHP version is PHP 7.0 or higher.
+- Nullable type declarations cannot be used until the minimum PHP version is PHP 7.1 or higher.
+- The `iterable` and `void` type declarations cannot be used until the minimum PHP version is PHP 7.1 or higher. The `void` can only be used as a return type.
+- The `object` type declaration cannot be used until the minimum PHP version is PHP 7.2 or higher.
+- Property type declarations cannot be used until the minimum PHP version is PHP 7.4 or higher.
+- `static` return type cannot be used until the minimum PHP version is PHP 8 or higher.
+- `mixed` type cannot be used until the minimum PHP version is PHP 8 or higher. Because `mixed` type includes `null`. it's not allowed to make it nullable.
+- Union types cannot be used until the minimum PHP version is PHP 8 or higher.
+- Intersection types cannot be used until the minimum PHP version is PHP 8.1 or higher.
+- `never` return type cannot be used until the minimum PHP version is PHP 8.1 or higher.
```
jrfnl, line 388:
```suggestion
- `never` (return type only) cannot be used until the minimum PHP version is PHP 8.1 or higher.
- Disjunctive normal form types (combining union types and intersection types) cannot be used until the minimum PHP version is PHP 8.2 or higher.
```

Might as well anticipate on PHP 8.2 as otherwise we'd need to edit this again in a few months time.

wordpress-coding-standards/php.md
```diff
@@ -370,6 +370,91 @@ $a = foo(
 );
 ```
 
+### Type declarations
+
+[alert]
```
jrfnl, line 375: Like with other new sections, I would move this part (the `alert` box) to the bottom of the section _after_ the CS rules (as these are not CS rules, but information about the PHP feature).

Also: is `alert` the red style ? Should this maybe be an info box (blue IIRC) instead ?

wordpress-coding-standards/php.md
```diff
@@ -370,6 +370,91 @@ $a = foo(
...
+- The `object` type declaration cannot be used until the minimum PHP version is PHP 7.2 or higher.
+- Property type declarations cannot be used until the minimum PHP version is PHP 7.4 or higher.
+- `static` return type cannot be used until the minimum PHP version is PHP 8 or higher.
+- `mixed` type cannot be used until the minimum PHP version is PHP 8 or higher. Because `mixed` type includes `null`. it's not allowed to make it nullable.
+- Union types cannot be used until the minimum PHP version is PHP 8 or higher.
+- Intersection types cannot be used until the minimum PHP version is PHP 8.1 or higher.
+- `never` return type cannot be used until the minimum PHP version is PHP 8.1 or higher.
+[/alert]
+
+Type declarations must have exactly one space before and after the type. The nullability operator (`?`) is regarded as part of the type declaration and there should be no space between this operator and the actual type. Class/interface name based type declarations should use the case of the class/interface name as declared, while the keyword-based type declarations should be lowercased.
```
jrfnl, line 391:
```suggestion
Type declarations must have exactly one space before and after the type. The nullability operator (`?`) is regarded as part of the type declaration and there should be no space between this operator and the actual type. Class/interface/enum name based type declarations should use the case of the class/interface/enum name as declared, while the keyword-based type declarations should be lowercased.
```

wordpress-coding-standards/php.md
```diff
@@ -370,6 +370,91 @@ $a = foo(
...
+
+Return type declarations should have no space between the closing parenthesis of the function declaration and the colon starting a return type.
+
+These rules apply to all structures allowing for type declarations: functions, closures, catch conditions as well as the PHP 7.4 arrow functions and typed properties.
```
jrfnl, line 395:
```suggestion
These rules apply to all structures allowing for type declarations: functions, closures, enums, catch conditions as well as the PHP 7.4 arrow functions and typed properties.
```

wordpress-coding-standards/php.md
```diff
@@ -370,6 +370,91 @@ $a = foo(
...
+
+The following examples showcase what type hints are currently available in the WordPress Core due to the minimum supported PHP version
```
jrfnl, line 397:
```suggestion
The following examples showcase what type declarations are currently available in WordPress Core due to the minimum supported PHP version:
```

To be honest, I'd be tempted to say: combine the two sets of examples. The info/alert boxes at the bottom will contain the info on what can be used in Core and what not.

wordpress-coding-standards/php.md
```diff
@@ -370,6 +370,91 @@ $a = foo(
...
+[warning]
+Adding type declarations to existing WordPress Core functions should be done with extreme care.
+[/warning]
+
+While rare, `array` and class/interface name based type declarations already exist in WordPress Core.
+
+The function signature of any function (method) which can be overloaded by plugins or themes should not be touched.
+This includes all existing `public` and `protected` class methods and any conditionally declared (pluggable) functions in the global namespace.
+
+This leaves, for now, only unconditionally declared functions in the global namespace, `private` class methods, and code newly being introduced as candidates for adding type declarations.
```
jrfnl, line 452:
```suggestion
This leaves, for now, only unconditionally declared functions in the global namespace, `private` class methods, and code  new to Core, as candidates for adding type declarations.
```

wordpress-coding-standards/php.md
```diff
@@ -370,6 +370,91 @@ $a = foo(
...
+_Usage in WordPress Core_
+
+[warning]
+Adding type declarations to existing WordPress Core functions should be done with extreme care.
+[/warning]
+
+While rare, `array` and class/interface name based type declarations already exist in WordPress Core.
+
+The function signature of any function (method) which can be overloaded by plugins or themes should not be touched.
+This includes all existing `public` and `protected` class methods and any conditionally declared (pluggable) functions in the global namespace.
```
jrfnl, line 450: I think this sentence can be left out.

wordpress-coding-standards/php.md
```diff
@@ -370,6 +370,91 @@ $a = foo(
...
+}
+```
+
+_Usage in WordPress Core_
+
+[warning]
+Adding type declarations to existing WordPress Core functions should be done with extreme care.
+[/warning]
+
+While rare, `array` and class/interface name based type declarations already exist in WordPress Core.
```
jrfnl, line 447: I think this sentence can be left out.

wordpress-coding-standards/php.md
```diff
@@ -370,6 +370,91 @@ $a = foo(
...
+
+The function signature of any function (method) which can be overloaded by plugins or themes should not be touched.
+This includes all existing `public` and `protected` class methods and any conditionally declared (pluggable) functions in the global namespace.
+
+This leaves, for now, only unconditionally declared functions in the global namespace, `private` class methods, and code newly being introduced as candidates for adding type declarations.
+
+For those, adding `callable`, class and interface name based parameter type declarations where the parameter type only expects one type and it would allow for simplification of existing/new code, is allowed and tentatively encouraged.
```
jrfnl, line 454: I think this sentence can be left out.

wordpress-coding-standards/php.md
```diff
@@ -370,6 +370,91 @@ $a = foo(
...
+
+Using the `array` keyword in type declarations is **strongly discouraged** for now, as most often, it would be better to use `iterable` to allow for more flexibility in the implementation and that keyword is not yet available for use in WordPress Core until the minimum requirements are raised to PHP 7.1.
```
jrfnl, line 456: If the "leave out" suggestions would be applied, I think this should be prefixed with "Note:".

```suggestion
Note: Using the `array` keyword in type declarations is **strongly discouraged** for now, as most often, it would be better to use `iterable` to allow for more flexibility in the implementation and that keyword is not yet available for use in WordPress Core until the minimum requirements are raised to PHP 7.1.
```

jrfnl (changes requested):
Gone through it with a fine toothcomb....

wordpress-coding-standards/php.md
jrfnl: Just checked - `alert` is the yellow style.

I think it may be good to have some standardization of when we use which type of box. What about this:
* `info` (blue) for information about PHP features and how to use them.
* `alert` (yellow) for information on whether or not something can be used in WP Core.
* `warning` (red) for anything really problematic (like using the `wp` namespace prefix).

wordpress-coding-standards/php.md
dingo-d: I think we can remove the examples for themes and plugins, to reduce any possible confusion. And as you've mentioned, people will have the info box with all the available types 🙂 👍🏼

wordpress-coding-standards/php.md
jrfnl: Oh and on that same note: what about standardizing the order as well ?

I'm thinking:
* CS rules
* Possible code samples
* Usage in Core (if needed)
* PHP feature info (if needed)

That would also imply that the "boxes" at the bottom of this section and for "trait use" would need to be reversed, but we'd be making the docs consistent and predictable for the reader if we have a set order for things.

jrfnl (commented):
See inline comments + the remark about the order of information.

wordpress-coding-standards/php.md
dingo-d: I'll open an issue and once we merge all the others I can open up a new PR with just these changes 👍🏼

jrfnl (approved):
Approving. There is a stray space in line 425 "code[space][space]new to Core", which I will remove before merging.

PR #104, opened 2022-08-20
The ::class constant

dingo-d:
This PR depends on #102. Once that is merged, this PR should be rebased to the master branch.

The PR adds rules about the `::class` constant and is the continuation of the additions of 'modern' PHP code in the WordPress PHP Coding Standards handbook based on the [make post](https://make.wordpress.org/core/2020/03/20/updating-the-coding-standards-for-modern-php/) by Juliette Reinders Folmer.

### DIFF
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -441,6 +441,18 @@ This leaves, for now, only unconditionally declared functions in the global name
 
 Note: Using the `array` keyword in type declarations is **strongly discouraged** for now, as most often, it would be better to use `iterable` to allow for more flexibility in the implementation and that keyword is not yet available for use in WordPress Core until the minimum requirements are raised to PHP 7.1.
 
+### The ::class constant
+
+When using the `::class` constant for class name resolution, the `class` keyword should be in lowercase and there should be no spaces around the `::` operator.
+
+```php
+// Correct.
+add_action( 'action_name', array( My_Class::class, 'method_name' ) );
+ 
+// Incorrect.
+add_action( 'action_name', array( My_Class :: CLASS, 'method_name' ) );
+```
+
 ## Declare Statements, Namespace, and Import Statements
 
 ### Namespace declarations


wordpress-coding-standards/php.md
```diff
@@ -366,6 +366,151 @@ $a = foo(
...
+
+This leaves, for now, only unconditionally declared functions in the global namespace, `private` class methods, and code newly being introduced as candidates for adding type declarations.
+
+For those, adding `callable`, class and interface name based parameter type declarations where the parameter type only expects one type and it would allow for simplification of existing/new code, is allowed and tentatively encouraged.
+
+Using the `array` keyword in type declarations is **strongly discouraged** for now, as most often, it would be better to use `iterable` to allow for more flexibility in the implementation and that keyword is not yet available for use in WordPress Core until the minimum requirements are raised to PHP 7.1.
+
+### The ::class constant
+
+When using the `::class` constant for class name resolution you should add no space between the `::class` and the preceding class name it applies to. There should be no space between the double colon and the `class` keyword, and the `class` keyword should be in lowercase.
```
jrfnl, line 499:
```suggestion
When using the `::class` constant for class name resolution, the `class` keyword should be in lowercase and there should be no spaces around the `::` operator.
```

wordpress-coding-standards/php.md
```diff
@@ -366,6 +366,151 @@ $a = foo(
...
+add_action( 'action_name', array( My_Class::class, 'method_name' ) );
+ 
+// Incorrect.
+add_action( 'action_name', array( My_Class :: CLASS, 'method_name' ) );
+```
+
+_Usage in WordPress Core_
+
+The `::class` constant can be used in WordPress Core.
+Replacing existing uses of `__CLASS__` with `self::class` and `get_called_class()` with `static::class`, where appropriate, is encouraged.
```
jrfnl, line 512: As there are no PHP cross-version issues/considerations, I don't think this needs to be mentioned in the coding standards docs.

Maybe we should just open up an issue in Trac for fixing these instead ? (if there isn't one open already) 

FYI: I'm actually coaching some people on writing a sniff to detect and auto-fix these. May still be a while before the sniff is available (in PHPCSExtra), but once it is, I'd suggest adding it to the `WordPress-Extra` ruleset.

If we wait a little with opening the Trac ticket, we could just auto-generate a patch instead of waiting for people to manually create patches ;-)

jrfnl (commented):
Left some pre-lim review notes.

wordpress-coding-standards/php.md
dingo-d: We can then just remove the entire Usage in WordPress Core part? I don't think mentioning that it can be used in the core has any merit - we wouldn't really mention a rule in the handbook if it cannot be used in the core in some form, no?

jrfnl: I agree with removing the "Usage in Core" part. We do mention some features in the CS docs which cannot be yet used in Core and for those having a section about "Usage in Core" makes sense. For things where there are no PHP cross-version considerations, I don't think we need to say that explicitly.

(In a way, that would be like saying: you can use arithmetic operators 😂 )

jrfnl:
Needs a rebase, but otherwise okay.

jrfnl (approved):
LGTM

PR #106, opened 2022-08-20
Instanceof and other operators

dingo-d:
Update the list in the spacing section and added some new code examples.

The PR adds rules about the additional operators and is the continuation of the additions of 'modern' PHP code in the WordPress PHP Coding Standards handbook based on the [make post](https://make.wordpress.org/core/2020/03/20/updating-the-coding-standards-for-modern-php/) by Juliette Reinders Folmer.

### DIFF
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -143,7 +143,7 @@ Where possible, dynamic values in tag names should also be as succinct and to th
 
 ### Space Usage
 
-Always put spaces after commas, and on both sides of logical, comparison, string and assignment operators.
+Always put spaces after commas, and on both sides of logical, arithmetic, comparison, string and assignment operators.
 
 ```php
 SOME_CONST === 23;
@@ -152,6 +152,8 @@ foo() && bar();
 array( 1, 2, 3 );
 $baz . '-5';
 $term .= 'X';
+if( $object instanceof Post_Type_Interface ) { [...] };
+$result = 2 ** 3; // 8.
 ```
 
 Put spaces on both sides of the opening and closing parentheses of control structure blocks.


jrfnl (commented):
I don't think this needs a separate section.

We already have the below in the handbook, so I think the phrasing may just need to be double-checked and if needs be, updated to cover these operators as well.

Maybe also add an example with one of the new (but available in PHP 5.6) operators to the pre-existing code sample ?

> ### Space Usage
> 
> Always put spaces after commas, and on both sides of logical, comparison, string and assignment operators.

=> I think `arithmetic` operators is missing here, other than that, I believe it already covers this rule.

wordpress-coding-standards/php.md
```diff
@@ -143,7 +143,7 @@ Where possible, dynamic values in tag names should also be as succinct and to th
 
 ### Space Usage
 
-Always put spaces after commas, and on both sides of logical, comparison, string and assignment operators.
+Always put spaces after commas, and on both sides of logical, arithmetic, comparison, string and assignment operators.
```
jrfnl, line 146: I just did a quick check [against the PHP manual](https://www.php.net/manual/en/language.operators.php) to verify if we now cover everything which should be covered here:
* Missing: [Bitwise Operators](https://www.php.net/manual/en/language.operators.bitwise.php), but then again they are so rarely used that I'm okay with not making this explicit. The sniff will check for them anyhow.

The following three types of operators identified in the manual do not fall under this rule, so are correctly omitted:
* [Error Control Operators](https://www.php.net/manual/en/language.operators.errorcontrol.php)
* [Execution Operators](https://www.php.net/manual/en/language.operators.execution.php) (I.e. the backticks)
* [Incrementing/Decrementing Operators](https://www.php.net/manual/en/language.operators.increment.php)

The manual also mentions [Array Operators](https://www.php.net/manual/en/language.operators.array.php). In practice those are already covered by the other categories, the reason why the PHP manual lists them separately is because those operators behave slightly differently for arrays.

Lastly, the manual mentions [Type Operators](https://www.php.net/manual/en/language.operators.type.php) (`instanceof`). To be fair, I think it's a bit weird that this is listed as a separate section in the PHP manual anyway, as effectively it is a comparison operator, though one which can only be used with objects.

wordpress-coding-standards/php.md
```diff
@@ -152,6 +152,9 @@ foo() && bar();
 array( 1, 2, 3 );
 $baz . '-5';
 $term .= 'X';
+$class = new My_Class();
```
jrfnl, line 155: This line seems redundant to me.

wordpress-coding-standards/php.md
```diff
@@ -152,6 +152,9 @@ foo() && bar();
...
+if( $class instanceof Post_Type_Interface ) { [...] };
```
jrfnl, line 156:
```suggestion
if( $object instanceof Post_Type_Interface ) { [...] };
```

jrfnl (approved):
LGTM

wordpress-coding-standards/php.md
```diff
@@ -152,6 +152,8 @@ foo() && bar();
 array( 1, 2, 3 );
 $baz . '-5';
 $term .= 'X';
+if( $object instanceof Post_Type_Interface ) { [...] };
```
GaryJones, line 155: The `if` should have had a space after it before the `(`?

dingo-d: Good catch, will fix it in a separate PR 👍🏼

PR #107, opened 2022-08-20
Visibility should always be declared

dingo-d:
This PR depends on #106. Once that is merged, this PR should be rebased to the master branch.

The PR adds rules about declaring visibility modifiers and is the continuation of the additions of 'modern' PHP code in the WordPress PHP Coding Standards handbook based on the [make post](https://make.wordpress.org/core/2020/03/20/updating-the-coding-standards-for-modern-php/) by Juliette Reinders Folmer.

### DIFF
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -552,6 +552,31 @@ class Foo {
 }
 ```
 
+### Visibility should always be declared
+
+For all constructs that allow it (properties, methods, class constants since PHP 7.1), visibility should be explicitly declared.
+Using the `var` keyword for property declarations is not allowed.
+
+```php
+// Correct.
+class Foo {
+    public $foo;
+
+    protected function bar() {}
+}
+
+// Incorrect.
+class Foo {
+    var $foo;
+
+    function bar() {}
+}
+```
+
+_Usage in WordPress Core_
+
+Visibility for class constants can not be used in WordPress Core until the minimum PHP version has been raised to PHP 7.1.
+
 ## Control Structures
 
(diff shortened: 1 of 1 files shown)


wordpress-coding-standards/php.md
```diff
@@ -432,6 +702,33 @@ class Example_Class { [...] }
...
+class Foo {
+    var $foo;
+
+    function bar() {}
+}
+```
+
+_Usage in WordPress Core_
+
+If any visibility modifiers are missing in WordPress Core code, they should be set to `public` so as not to break backward-compatibility.
```
jrfnl, line 728: I would remove this line - this will be fixed in the same patch as the update to WPCS 3.0.0, so after that, this shouldn't be an issue anymore.

jrfnl (commented):
Just one small remark.

wordpress-coding-standards/php.md
```diff
@@ -552,6 +552,31 @@ class Foo {
...
+class Foo {
+    var $foo;
+
+    function bar() {}
+}
+```
+
+_Usage in WordPress Core_
+
+Visibility for class constants can not be used in WordPress Core until the minimum PHP version has been raised to PHP 7.1.
```
jrfnl, line 578: Just wondering: should this be an `alert` box ?

dingo-d: Other WP Core usages are currently not in the alert box (for instance import use statements one).

I'd leave this as is for now, and we can then see later if we'd like to wrap all of them in a box.

jrfnl (approved):
LGTM

PR #108, opened 2022-08-20
Property and method modifier order

dingo-d:
This PR depends on #107. Once that is merged, this PR should be rebased to the master branch.

The PR adds rules about the order of property and method modifiers and is the continuation of the additions of 'modern' PHP code in the WordPress PHP Coding Standards handbook based on the [make post](https://make.wordpress.org/core/2020/03/20/updating-the-coding-standards-for-modern-php/) by Juliette Reinders Folmer.

### DIFF
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -710,6 +710,90 @@ _Usage in WordPress Core_
 
 Visibility for class constants can not be used in WordPress Core until the minimum PHP version has been raised to PHP 7.1.
 
+### Visibility and modifier order
+
+When using multiple modifiers for a class declaration, the order should be as follows:
+
+1. First the optional `abstract` or `final` modifier.
+2. Next, the optional `readonly` modifier.
+
+```php
+// Correct.
+abstract readonly class Foo {}
+
+// Incorrect.
+readonly final class Foo {}
+```
+
+When using multiple modifiers for a constant declaration inside object-oriented structures, the order should be as follows:
+
+1. First the optional `final` modifier.
+2. Next, the visibility modifier.
+
+```php
+// Correct.
+class Foo {
+    private const LABEL = 'Book';
+
+    final public const SLUG = 'book';
+}
+
+// Incorrect.
+interface Mailer {
+    protected final const SLUG = 'book';
+}
+```
+
+When using multiple modifiers for a property declaration, the order should be as follows:
+
+1. First a visibility modifier.
+2. Next, the optional `static` or `readonly` modifier (these keywords are mutually exclusive).
+3. Finally, the `type` declaration (if possible due to PHP version).
+
+```php
+// Correct.
+abstract class Foo {
+    public static $foo;
+
+    private readonly string $bar;
+}
+
+// Incorrect.
+class Foo {
+    static public $foo;
+
+    readonly private string $foo;
+}
+```
+
+When using multiple modifiers for a method declaration, the order should be as follows:
+
+1. First the optional `abstract` or `final` modifier.
+2. Then, a visibility modifier.
+3. Finally, the optional `static` modifier.
+
+```php
+// Correct.
+abstract class Foo {
+    abstract protected static function bar();
+}
+
+// Incorrect.
+class Foo {
+    static protected final function bar() {
+        // Code.
+    };
+}
+```
+
+[info]
+Visibility for OO constants can be declared since PHP 7.1.
+Typed properties are available since PHP 7.4.
+Readonly properties are available since PHP 8.1.
+`final` modifier for constants in object-oriented structures is available since PHP 8.1.
+Readonly classes are available since PHP 8.2.
+[/info]
+
 ## Control Structures
 
 ### Use `elseif`, not `else if`
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -726,7 +726,7 @@ When using multiple modifiers for a _property declaration_, the order should be 
 
 1. First a visibility modifier.
 2. Next, the optional `static` or `readonly` modifier (these keywords are mutually exclusive).
-3. Finally, the `type` declaration (if possible due to PHP version).
+3. Finally, the optional `type` declaration.
 
 When using multiple modifiers for a _method declaration_, the order should be as follows:
 
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -712,61 +712,23 @@ Visibility for class constants can not be used in WordPress Core until the minim
 
 ### Visibility and modifier order
 
-When using multiple modifiers for a class declaration, the order should be as follows:
+When using multiple modifiers for a _class declaration_, the order should be as follows:
 
 1. First the optional `abstract` or `final` modifier.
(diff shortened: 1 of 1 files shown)


jrfnl (commented):
Should we mention the `readonly` keyword (PHP 8.1 properties) and property type declarations (PHP 7.4) somewhere in this section ? Or leave it for now until there is sight of a WP minimum PHP version bump ?

Or maybe we should split this into two paragraphs ?
* Properties (visibility, static, readonly, type)
* Methods (abstract/final, visibility, static)

dingo-d:
If we add `readonly` and we split this to properties and methods, should we have a paragraph for readonly classes as well?

So we'd have:

- Classes (abstract/final, readonly)
- Properties (visibility, static, readonly, type)
- Methods (abstract/final, visibility, static)

?

GaryJones:
I think this needs rebasing to be easier to review.

dingo-d:
@GaryJones I've rebased it, should be easier to review now 🙂

jrfnl:
> If we add readonly and we split this to properties and methods, should we have a paragraph for `readonly` classes as well?

We just discussed this and I agree, I think having a separate paragraph for each would be good. Though IMO that would mean four paragraphs:
* Classes (abstract/final, readonly) => mind: enums also have a type (but that goes _after_ the `enum` keyword.
* Constants (visibility) - not that interesting yet as there is only one type of modifier allowed, but there has been talk about allowing type declarations there too
* Properties (visibility, static, readonly, type) - note: `static`/`readonly` are currently mutually exclusive.
* Methods (abstract/final, visibility, static)

wordpress-coding-standards/php.md
```diff
@@ -710,6 +710,92 @@ _Usage in WordPress Core_
...
+    protected final const SLUG = 'book';
+}
+```
+
+When using multiple modifiers for a property declaration, the order should be as follows:
+
+1. First the optional `abstract` modifier.
+2. Then, a visibility modifier.
+3. Next, the optional `static` or `readonly` modifiers (they are currently mutually exclusive).
+4. Finally, the `type` declaration (if possible due to PHP version).
```
jrfnl, line 755:
```suggestion
4. Finally, the `type` declaration (PHP 7.4+.
```

wordpress-coding-standards/php.md
```diff
@@ -710,6 +710,92 @@ _Usage in WordPress Core_
...
+    public static $foo;
+
+    private readonly string $bar;
+}
+
+// Incorrect.
+abstract class Foo {
+    static public $foo;
+
+    private abstract string $foo;
```
jrfnl, line 769: Parse error...

wordpress-coding-standards/php.md
```diff
@@ -710,6 +710,92 @@ _Usage in WordPress Core_
...
+3. Finally, the optional `static` modifier.
+
+```php
+// Correct.
+abstract class Foo {
+    abstract protected static function bar();
+}
+
+// Incorrect.
+abstract class Foo {
```
jrfnl, line 786:
```suggestion
class Foo {
```

wordpress-coding-standards/php.md
```diff
@@ -710,6 +710,92 @@ _Usage in WordPress Core_
...
+        // Code.
+    };
+}
+```
+
+[info]
+Readonly properties are available since PHP 8.1.
+`readonly` modifier for constants in object-oriented structures is available since PHP 8.1.
+Readonly classes are available since PHP 8.2.
+[/info]
```
jrfnl, line 797:
```suggestion
[info]
Visibility for OO constants can be declared since PHP 7.1.
Typed properties are available since PHP 7.4.
Readonly properties are available since PHP 8.1.
`final` modifier for constants in object-oriented structures is available since PHP 8.1.
Readonly classes are available since PHP 8.2.
[/info]
```

jrfnl (commented):
Reviewed. I hope I caught everything.

Not sure we'd want to have the `PHP #.#+` markings inline with the actual order rules, but there were some "if available"s, so either do it with all or don't do it with any of them and just let it be noted in the PHP info block.

wordpress-coding-standards/php.md
```diff
@@ -710,6 +710,92 @@ _Usage in WordPress Core_
...
+
+// Incorrect.
+interface Mailer {
+    protected final const SLUG = 'book';
+}
+```
+
+When using multiple modifiers for a property declaration, the order should be as follows:
+
+1. First the optional `abstract` modifier.
```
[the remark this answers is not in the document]
dingo-d: Correct, this was left over from the initial copy, my mistake 🙈

wordpress-coding-standards/php.md
```diff
@@ -710,6 +710,92 @@ _Usage in WordPress Core_
...
+
+When using multiple modifiers for a constant declaration inside object-oriented structures, the order should be as follows:
+
+1. First the optional `final` modifier.
+2. Finally, the visibility modifier.
+
+```php
+// Correct.
+class Foo {
+    private const LABEL = 'Book';
```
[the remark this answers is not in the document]
dingo-d: I'd leave this as is, because I don't think we are using such class names in other code examples.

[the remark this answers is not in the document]
GaryJones: Agreed - keep Foo and the comments.

GaryJones (commented):
I wonder if this could all be simplified down as per https://www.php-fig.org/psr/psr-12/#46-abstract-final-and-static, which says:

> When present, the `abstract` and `final` declarations MUST precede the visibility declaration.
>
> When present, the `static` declaration MUST come after the visibility declaration.

We'd need to add in `readonly`, and types, but it seems that it could then be covered in just one or two code snippets.

dingo-d:
I'm fine with having one code example covering all the things, we opted for splitting it into several paragraphs for easier maintenance.

dingo-d:
@GaryJones I've made some changes based on the discussion I've had with Juliette:

- Left the specific parts about the order for certain OO constructs
- Create only one code example with correct and incorrect usage
- Added bullets to the info block (we have a similar info block for the type declaration)

Does this seem a bit more manageable?

wordpress-coding-standards/php.md
```diff
@@ -710,6 +710,62 @@ _Usage in WordPress Core_
...
+When using multiple modifiers for a _constant declaration_ inside object-oriented structures, the order should be as follows:
+
+1. First the optional `final` modifier.
+2. Next, the visibility modifier.
+
+When using multiple modifiers for a _property declaration_, the order should be as follows:
+
+1. First a visibility modifier.
+2. Next, the optional `static` or `readonly` modifier (these keywords are mutually exclusive).
+3. Finally, the `type` declaration (if possible due to PHP version).
```
jrfnl, line 729:
```suggestion
3. Finally, the optional `type` declaration.
```

I wouldn't mention the "(if possible due to PHP version)" as that applies to half the things we mention in this section and we don't mention it elsewhere. Let people refer to the info block for that.

jrfnl (commented):
Just one small comment, other than that: LGTM!

jrfnl (approved):

GaryJones (approved):
✅

PR #109, opened 2022-08-20
Object instantiation

dingo-d:
This PR depends on #108. Once that is merged, this PR should be rebased to the master branch.

The PR adds rules about the object instantiation and is the continuation of the additions of 'modern' PHP code in the WordPress PHP Coding Standards handbook based on the [make post](https://make.wordpress.org/core/2020/03/20/updating-the-coding-standards-for-modern-php/) by Juliette Reinders Folmer.

### DIFF
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -766,6 +766,22 @@ class Foo {
 - Readonly classes are available since PHP 8.2.
 [/info]
 
+### Object Instantiation
+
+When instantiating a new object instance, parenthesis must always be used, even when not strictly necessary.
+There should be no space between the name of the class being instantiated and the opening parenthesis.
+
+```php
+// Correct.
+$foo = new Foo();
+$anonymous_class = new class( $parameter ) { ... };
+$instance = new static();
+
+// Incorrect.
+$foo = new Foo;
+$anonymous_class = new class ( $input ) { ... };
+```
+
 ## Control Structures
 
 ### Use `elseif`, not `else if`


wordpress-coding-standards/php.md
```diff
@@ -432,6 +702,76 @@ class Example_Class { [...] }
...
+        // Code.
+    };
+}
+```
+
+### Object instantiation
+
+When instantiating a new object instance, parenthesis must always be used, even when not strictly necessary.
+There should be no space between the name of the class being instantiated and the opening parenthesis.
+Assigning the return value of an object instantiation by reference is not allowed (new by reference has not been supported by PHP for quite a long while now).
```
jrfnl, line 762: I would leave this out, it's a PHP error after all.

jrfnl (commented):
Just one small remark.

jrfnl (approved):
LGTM. ✅ 

Might it be an idea to have at least one of the object instantiations for "Correct" pass one of more variables ?
(obviously doesn't apply to "Incorrect")

dingo-d:
@jrfnl like this?

```php
// Correct.
$foo = new Foo();
$anonymous_class = new class( $dependency ) { ... };
$instance = new static();

// Incorrect.
$foo = & new Foo;
$foo = new Foo ();
```

jrfnl:
@dingo-d Actually, I realized that for the whitespace issue in "Incorrect" a param could also be used.

As for the code change, yes, something like that, though maybe keep the param name more "generalized" ? i.e. `$param`, `$bar`, `$input` or something.

dingo-d:
Last sanity check

```php
// Correct.
$foo = new Foo();
$anonymous_class = new class( $parameter ) { ... };
$instance = new static();

// Incorrect.
$foo = & new Foo; // Assign by reference is not allowed for object instantiation.
$foo = new Foo ();
$anonymous_class = new class ($input) { ... };
```

The parameter in the incorrect example is bothering me a bit because the sniffs for object instantiation are not checking for whitespace inside the parenthesis, so those are two different sniffs. Or did you mean, adding a parameter, but with correct spacing inside the parenthesis (but it's still incorrect because the space before the opening parenthesis)?

jrfnl (approved):

wordpress-coding-standards/php.md
```diff
@@ -766,6 +766,23 @@ class Foo {
 - Readonly classes are available since PHP 8.2.
 [/info]
 
+### Object Instantiation
+
+When instantiating a new object instance, parenthesis must always be used, even when not strictly necessary.
```
GaryJones, line 771:
```suggestion
When instantiating a new object instance, parenthesis must always be used.
```

The presence of "must" makes the final part of the sentence redundant. It could be left to remove ambiguity or removed to keep things simple and direct and not even entertain the idea that there's a scenario when they may not be necessary.

wordpress-coding-standards/php.md
```diff
@@ -766,6 +766,23 @@ class Foo {
...
+There should be no space between the name of the class being instantiated and the opening parenthesis.
+
+```php
+// Correct.
+$foo = new Foo();
+$anonymous_class = new class( $parameter ) { ... };
+$instance = new static();
+
+// Incorrect.
+$foo = & new Foo; // Assign by reference is not allowed for object instantiation.
```
GaryJones, line 781: > Assign by reference is not allowed for object instantiation.

Is that a WPCS decision or a PHP technical thing?

dingo-d: It's a parse error in PHP: https://3v4l.org/b8vrf

GaryJones: OK, so why are we highlighting ways to break PHP in our incorrect examples? There are infinite ways to create parse errors. I'd rather see the incorrect example focus on the WPCS aspect (i.e. use of paratheses, spacing around / inside / after parentheses, etc.).

dingo-d: It was mentioned in the original make post: https://make.wordpress.org/core/2020/03/20/updating-the-coding-standards-for-modern-php/

So I added it. I can remove it (makes sense not to have parse errors in the code examples).

I'll just remove the `&` in this example 👍🏼

GaryJones (approved):
✅

PR #110, opened 2022-08-20
Function closing brace

dingo-d:
This PR depends on #109. Once that is merged, this PR should be rebased to the master branch.

The PR adds rules about the function closing brace and is the continuation of the additions of 'modern' PHP code in the WordPress PHP Coding Standards handbook based on the [make post](https://make.wordpress.org/core/2020/03/20/updating-the-coding-standards-for-modern-php/) by Juliette Reinders Folmer.

### DIFF
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -286,6 +286,8 @@ switch ( $type ) {
 
 Remove trailing whitespace at the end of each line. Omitting the closing PHP tag at the end of a file is preferred. If you use the tag, make sure you remove trailing whitespace.
 
+There should be no trailing blank lines at the end of a function body.
+
 ## Formatting
 
 ### Brace Style


jrfnl (commented):
I wonder if this should have an explicit separate section or if we could add this to one of the pre-existing sections.

Looking at the pre-existing sections, there are a couple of potential places where this could be added:
* **Space usage** - though feels a bit odd as it would be the only one regarding blank lines.
* **Remove Trailing Spaces** - in this case, it's a about trailing new lines, but not necessarily a bad match.
* **Brace style** - doesn't feel right as the section is more about control structures with optional braces, while this rule is about function declarations. Adding it there _could_ give the impression that the rule is not just about function declarations.

So, "Remove trailing spaces" seems the most appropriate one.

An extra paragraph there stating "there should be no trailing blank lines at the end of a function body" should suffice IMO.

I also think we can leave out the code sample as it's pretty clear what that means.

wordpress-coding-standards/php.md
```diff
@@ -286,6 +286,8 @@ switch ( $type ) {
 
 Remove trailing whitespace at the end of each line. Omitting the closing PHP tag at the end of a file is preferred. If you use the tag, make sure you remove trailing whitespace.
 
+There should be no trailing blank lines at the end of a function body as well.
```
jrfnl, line 289:
```suggestion
There should be no trailing blank lines at the end of a function body.
```

jrfnl (approved):
LGTM

PR #111, opened 2022-08-20
Include require statements

dingo-d:
This PR depends on #110. Once that is merged, this PR should be rebased to the master branch.

The PR adds rules about the include/require statements and is the continuation of the additions of 'modern' PHP code in the WordPress PHP Coding Standards handbook based on the [make post](https://make.wordpress.org/core/2020/03/20/updating-the-coding-standards-for-modern-php/) by Juliette Reinders Folmer.

### DIFF
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -78,6 +78,21 @@ echo "<a href='{$escaped_link}'>text with a ' single quote</a>";
 
 Text that goes into HTML or XML attributes should be escaped so that single or double quotes do not end the attribute value and invalidate the HTML, causing a security issue. See [Data Validation](https://developer.wordpress.org/plugins/security/data-validation/) in the Plugin Handbook for further details.
 
+### Writing include/require statements
+
+Because `include[_once]` and `require[_once]` are language constructs, they do not need parentheses around the path, so those shouldn't be used. There should only be one space between the path and the include/require keywords.
+
+It is _strongly recommended_ to use `require[_once]` for unconditional includes. When using `include[_once]`, PHP will throw a warning when the file is not found but will continue execution, which will almost certainly lead to other errors/warnings/notices being thrown if your application depends on the file loaded, potentially leading to security leaks. For that reason, `require[_once]` is generally the better choice as it will throw a `Fatal Error` if the file cannot be found.
+
+```php
+// Correct.
+require_once ABSPATH . 'file-name.php';
+
+// Incorrect.
+include_once  ( ABSPATH . 'file-name.php' );
+require_once     __DIR__ . '/file-name.php';
+```
+
 ## Naming
 
 ### Naming Conventions


wordpress-coding-standards/php.md
```diff
@@ -78,6 +78,21 @@ echo "<a href='{$escaped_link}'>text with a ' single quote</a>";
 
 Text that goes into HTML or XML attributes should be escaped so that single or double quotes do not end the attribute value and invalidate the HTML, causing a security issue. See [Data Validation](https://developer.wordpress.org/plugins/security/data-validation/) in the Plugin Handbook for further details.
 
+### Writing include/require statements
+
+Because `include[_once]` and `require[_once]` are language constructs, they do not need parentheses around the path. There should only be one space between the path and the include/require keywords.
+It is _strongly recommended_ to use `require[_once]` for unconditional includes.
```
jrfnl, line 84: Suggestion: add a blank line before this line as the first line is about CS, while the second line and after is about recommended "best practices".

wordpress-coding-standards/php.md
```diff
@@ -78,6 +78,21 @@ echo "<a href='{$escaped_link}'>text with a ' single quote</a>";
...
+When using `include[_once]` PHP will throw a warning when the file is not found, but will continue execution which will almost certainly lead to other errors/warnings/notices being thrown if your application depends on the file being available, potentially leading to security leaks. For that reason, `require[_once]` is generally the better choice as it will throw a `Fatal Error` if the file cannot be found.
```
jrfnl, line 85:
```suggestion
When using `include[_once]` PHP will throw a warning when the file is not found, but will continue execution which will almost certainly lead to other errors/warnings/notices being thrown if your application depends on the file having been loaded, potentially leading to security leaks. For that reason, `require[_once]` is generally the better choice as it will throw a `Fatal Error` if the file cannot be found.
```

wordpress-coding-standards/php.md
```diff
@@ -78,6 +78,21 @@ echo "<a href='{$escaped_link}'>text with a ' single quote</a>";
 
 Text that goes into HTML or XML attributes should be escaped so that single or double quotes do not end the attribute value and invalidate the HTML, causing a security issue. See [Data Validation](https://developer.wordpress.org/plugins/security/data-validation/) in the Plugin Handbook for further details.
 
+### Writing include/require statements
+
+Because `include[_once]` and `require[_once]` are language constructs, they do not need parentheses around the path. There should only be one space between the path and the include/require keywords.
```
jrfnl, line 83:
```suggestion
Because `include[_once]` and `require[_once]` are language constructs, they do not need parentheses around the path. There should only be one space between the include/require keyword and the path.
```

I'm not fully convinced about the first sentence. It feels a bit like a soft recommendation, while it was proposed as a hard rule.

Re: the tweak in the second sentence - IMO this reads more natural as that's the order of in the code.

wordpress-coding-standards/php.md
```diff
@@ -78,6 +78,21 @@ echo "<a href='{$escaped_link}'>text with a ' single quote</a>";
...
+It is _strongly recommended_ to use `require[_once]` for unconditional includes.
+When using `include[_once]` PHP will throw a warning when the file is not found, but will continue execution which will almost certainly lead to other errors/warnings/notices being thrown if your application depends on the file being available, potentially leading to security leaks. For that reason, `require[_once]` is generally the better choice as it will throw a `Fatal Error` if the file cannot be found.
+
+```php
+// Correct.
+require_once ABSPATH . 'file-name.php';
+
+// Incorrect.
+include_once  ( ABSPATH . 'file-name.php' );
+require_once     __FILE__ . '/file-name.php';
```
jrfnl, line 93:
```suggestion
require_once     __DIR__ . '/file-name.php';
```

jrfnl (changes requested):
Left some inline notes.

wordpress-coding-standards/php.md
dingo-d: > Because `include[_once]` and `require[_once]` are language constructs, you mustn't add parentheses around the path.

This reads as a hard rule, the only issue is that it kinda reads like it would cause a fatal error if you add it (which it won't) 🤷🏼‍♂️

jrfnl: > Because include[_once] and require[_once] are language constructs, you mustn't add parentheses around the path.

What about using "shouldn't" instead of "mustn't" ?

Or even:
> Because `include[_once]` and `require[_once]` are language constructs, they do not need parentheses around the path, so those shouldn't be used.

wordpress-coding-standards/php.md
```diff
@@ -78,6 +78,22 @@ echo "<a href='{$escaped_link}'>text with a ' single quote</a>";
 
 Text that goes into HTML or XML attributes should be escaped so that single or double quotes do not end the attribute value and invalidate the HTML, causing a security issue. See [Data Validation](https://developer.wordpress.org/plugins/security/data-validation/) in the Plugin Handbook for further details.
 
+### Writing include/require statements
+
+Because `include[_once]` and `require[_once]` are language constructs, they do not need parentheses around the path, so those shouldn't be used. There should only be one space between the path and the include/require keywords.
+
+It is _strongly recommended_ to use `require[_once]` for unconditional includes.
+When using `include[_once]` PHP will throw a warning when the file is not found, but will continue execution which will almost certainly lead to other errors/warnings/notices being thrown if your application depends on the file having been loaded, potentially leading to security leaks. For that reason, `require[_once]` is generally the better choice as it will throw a `Fatal Error` if the file cannot be found.
```
GaryJones, line 86:
```suggestion
Because `include[_once]` and `require[_once]` are language constructs, they do not need parentheses around the path, so those shouldn't be used. There should only be one space between the path and the include/require keywords.

It is _strongly recommended_ to use `require[_once]` for unconditional includes. When using `include[_once]`, PHP will throw a warning when the file is not found but will continue execution, which will almost certainly lead to other errors/warnings/notices being thrown if your application depends on the file loaded, potentially leading to security leaks. For that reason, `require[_once]` is generally the better choice as it will throw a `Fatal Error` if the file cannot be found.
```

(mostly just comma changes)

GaryJones (approved):
✅

dingo-d:
@jrfnl Mind giving this another check, and merging if it's ok?

jrfnl (approved):
LGTM! Thanks @dingo-d !

PR #112, opened 2022-08-20
Increment decrement operators

dingo-d:
This PR depends on #111. Once that is merged, this PR should be rebased to the master branch.

The PR adds rules about the increment/decrement operators and is the continuation of the additions of 'modern' PHP code in the WordPress PHP Coding Standards handbook based on the [make post](https://make.wordpress.org/core/2020/03/20/updating-the-coding-standards-for-modern-php/) by Juliette Reinders Folmer.

### DIFF
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -227,6 +227,17 @@ if ( $foo && ( $bar || $baz ) ) { ...
 my_function( ( $x - 1 ) * 5, $y );
 ```
 
+When using increment (`++`) or decrement (`--`) operators, there should be no spaces between the operator and the variable it applies to.
+
+```php
+// Correct.
+for ( $i = 0; $i < 10; $i++ ) { [...] }
+
+// Incorrect.
+for ( $i = 0; $i < 10; $i ++ ) { [...] }
+++   $b; // Multiple spaces.
+```
+
 ### Indentation
 
 Your indentation should always reflect logical structure. Use **real tabs**, **not spaces**, as this allows the most flexibility across clients.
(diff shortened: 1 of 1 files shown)


jrfnl (commented):
I think the "spacing" part should be moved to the "Space usage" section. (including the code samples)

For the recommended use of pre vs post, I do think having a separate subsection is the right way to go and the subsection is in the right place.

wordpress-coding-standards/php.md
```diff
@@ -227,6 +227,17 @@ if ( $foo && ( $bar || $baz ) ) { ...
 my_function( ( $x - 1 ) * 5, $y );
 ```
 
+When writing increment (`++`) or decrement (`--`) operators, there should be no spaces between the operator and the variable it applies to.
```
jrfnl, line 230:
```suggestion
When using increment (`++`) or decrement (`--`) operators, there should be no spaces between the operator and the variable it applies to.
```

jrfnl (approved):
LGTM

wordpress-coding-standards/php.md
```diff
@@ -600,6 +611,21 @@ While this operator does exist in Core, it is often used lazily instead of doing
...
+Pre-increment/decrement should be favoured over post-increment/decrement for stand-alone statements.
+
+Pre-increment/decrement will increment/decrement and then return, while post-increment/decrement will return and then increment/decrement.
+Using the "pre" version is slightly more performant and can prevent future bugs when code gets moved around.
+
+```php
+// Correct: Pre-decrement.
+--$a;
+
+// Incorrect: Post-decrement.
```
GaryJones, line 625: Rather than "Correct" and "Incorrect", I might have gone with "Recommended" and "Discouraged" since the overall rule seems less strict ("favoured").

wordpress-coding-standards/php.md
```diff
@@ -600,6 +611,21 @@ While this operator does exist in Core, it is often used lazily instead of doing
 
 > Warning: Prior to PHP 8.0.0, it was possible for the @ operator to disable critical errors that will terminate script execution. For example, prepending @ to a call of a function that did not exist, by being unavailable or mistyped, would cause the script to terminate with no indication as to why.
 
+### Increment/decrement operators
+
+Pre-increment/decrement should be favoured over post-increment/decrement for stand-alone statements.
```
GaryJones, line 616: Instead of "favoured", I would have gone with "preferred", since that avoids the en_US / en_GB (and others) difference in the spelling of "favo[u]red".

wordpress-coding-standards/php.md
dingo-d: Everywhere else we are using the `Correct/Incorrect` syntax, and I think that Juliette mentioned this will get reported as incorrect usage in WPCS 3.0 (if I recall correctly).

jrfnl: Yes, there is a sniff which will start flagging this, but as the original text used "favoured" (strong recommendation), in my WIP branch, I've got that sniff throwing a `warning` for this code, not an `error`.

In that sense, we could choose to (consistently) used "Correct/Incorrect" for everything that will be throwing errors and "Recommended/Discouraged" for everything that will be throwing warnings ?

The `Core` ruleset is mostly errors, but there are some things set to warning. We could cross-check the "final" settings in WPCS with the docs before publishing WPCS 3.0.0 ?

PR #116, opened 2022-09-05
Fix code example

dingo-d:
A fix after a comment by @GaryJones in https://github.com/WordPress/wpcs-docs/pull/106#discussion_r962182050_

### DIFF
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -85,18 +85,18 @@ Text that goes into HTML or XML attributes should be escaped so that single or d
 Use lowercase letters in variable, action/filter, and function names (never `camelCase`). Separate words via underscores. Don't abbreviate variable names unnecessarily; let the code be unambiguous and self-documenting.
 
 ```php
-function some_name( $some_variable ) { [...] }
+function some_name( $some_variable ) {}
 ```
 
 Class, trait, interface and enum names should use capitalized words separated by underscores. Any acronyms should be all upper case.
 
 ```php
-class Walker_Category extends Walker { [...] }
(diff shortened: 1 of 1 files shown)


GaryJones (approved):

jrfnl (approved):

jrfnl:
While looking at this line again now... one thing sprang to mind (reason I've not merged this yet):

I believe the `[...]` in the code sample intends to indicate "code here".
However, the square brackets in that bit `[...]` (which I believe come from typical IDEs using them), could also be interpreted as short array/short list brackets.... (by people not using IDEs/using a different IDE).

Considering that that **_always_** comes up as a discussion point again, I wonder if we should change that part of the code sample from `{ [...] };` to `{}`. What do you think ?

(Note: includes removing the redundant semi-colon after the end bracket, which would get flagged by WPCS as an empty statement otherwise)

dingo-d:
I've fixed the code samples as suggested and merged this (it's a small fix). Thanks for the suggestion 👍🏼

wordpress-coding-standards/php.md
```diff
@@ -152,7 +152,7 @@ foo() && bar();
 array( 1, 2, 3 );
 $baz . '-5';
 $term .= 'X';
-if( $object instanceof Post_Type_Interface ) { [...] };
+if ( $object instanceof Post_Type_Interface ) {};
```
jrfnl, line 155: @dingo, thanks for updating this, though this line still has a stray `;` after the closing curly....

dingo-d: 😱 Doh! Will fix this asap 🙈

PR #119, opened 2022-10-28
PHP docs: remove a duplication

jrfnl:
This same rule is also mentioned in the type declaration section (second paragrah).

Ref: https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#type-declarations


GaryJones (approved):
✅

dingo-d (approved):

PR #120, opened 2022-10-28
PHPdocs: discourage use of reserved keywords for parameter names

jrfnl:
Most of this has already been fixed in WP Core as part of making Core compatible with PHP 8.0 and it would be great if no new instances of this issue would be introduced.

Ref:
* https://core.trac.wordpress.org/ticket/51553
* https://core.trac.wordpress.org/ticket/55327
* https://core.trac.wordpress.org/ticket/55650
* https://core.trac.wordpress.org/ticket/56788


GaryJones (approved):
✅

dingo-d (approved):

PR #123, opened 2022-12-12
PHP: adjust the file name rule

jrfnl:
The three exceptions which were previously listed have recently been fixed in Core, so the exception should be removed.

At the same time, we should make the exception for test files explicit with an eye on the upstream changes in PHPUnit.

Refs:
* https://core.trac.wordpress.org/changeset/54254
* https://core.trac.wordpress.org/ticket/53010
* WordPress/WordPress-Coding-Standards#1995
* sebastianbergmann/phpunit#4109

### DIFF
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -135,7 +135,8 @@ Class file names should be based on the class name with `class-` prepended and t
 class-wp-error.php
 ```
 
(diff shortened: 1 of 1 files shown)


SergeyBiryukov (approved):

dingo-d (approved):

Dirtylo911:
Jrfnl

PR #126, opened 2023-02-28
Replace Broken Link

SavPhill:
The link on line 222 is broken. I have replaced it with the correct link to https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#4-hooks-actions-and-filters


GaryJones (approved):
✅

PR #128, opened 2023-06-14
Improved Readability

iamibrahimriaz:
Added Good Practice & Bad Practice instead of Good & Bad.


dingo-d (approved):

jrfnl (commented):
I'm not a fan of this change.

1. It makes the page even more inconsistent with the rest of the coding standards pages which mostly use "Correct" and "Incorrect".
2. In most cases, these are hard rules and not suggestions, so the phrasing "Good practice"/"Bad practice" does not apply.

Additionally, the punctuation of the adjusted comments does not comply with the docs coding standards.

PR #132, opened 2023-07-22
Update wpdb::prepare() placeholder notes

### LINKED ISSUE #103: WPDB to support %i in 6.1
The [Formatting SQL Statements](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#formatting-sql-statements) section currently includes a paragraph on [string and integer placeholders](https://github.com/WordPress/wpcs-docs/blob/dff1d8622d201bf1d9bc3b9896f783a44778b0cc/wordpress-coding-standards/php.md?plain=1#L465):

> `%s` is used for string placeholders and `%d` is used for integer placeholders. Note that they are not 'quoted'! `$wpdb->prepare()` will take care of escaping and quoting for us. The benefit of this is that it is easy to see at a glance whether something has been escaped or not because it happens right when the query happens.

Is it worth adding that (when WordPress 6.1 is available) developers will be able to use `%i` for Identifiers (e.g. field and table names)?

Maybe this should be a separate paragraph? and maybe only added when [6.1 is released in October](https://make.wordpress.org/core/2022/06/23/wordpress-6-1-planning-roundup/)?

---

The intention is to avoid the risky feature where WordPress does not quote string/integer/float placeholders when they contain argnum/format, something that can be used today when (incorrectly) trying to use a variable for field/table names, e.g.

```php
$field = 'bad`quoting';
$value = '5 AND 1=1';

$wpdb->prepare( 'WHERE `%1$s` = %2$s', $field, $value ); // WHERE `bad`quoting` = 5 AND 1=1
```

This relates to WordPress [Ticket #52506](https://core.trac.wordpress.org/ticket/52506), [Changeset #53575](https://core.trac.wordpress.org/changeset/53575), and [Pull Request #2072](https://github.com/WordPress/WordPress-Coding-Standards/pull/2072) for WordPress-Coding-Standards (thanks to Juliette, who has done a nice writeup about it, and pointed me to this repo).

craigfrancis:
Ref [Issue 103](https://github.com/WordPress/wpcs-docs/issues/103)

Closes #103

### DIFF
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -873,7 +873,7 @@ The following placeholders can be used in the query string:
 - %s (string)
 - %i (identifier, e.g. table/field names)
 
-`$wpdb->prepare()` will take care of escaping and quoting for us. This makes it easy to see at a glance whether something has been escaped or not because it happens right when the query happens.
+Note that these placeholders should not be 'quoted'! `$wpdb->prepare()` will take care of escaping and quoting for us. This makes it easy to see at a glance whether something has been escaped or not because it happens right when the query happens.
 
 See [Data Validation](https://developer.wordpress.org/plugins/security/data-validation/) in the Plugin Handbook for further details.
 
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -866,7 +866,14 @@ $id = some_foo_number(); // Data we expect to be an integer, but we're not certa
 $wpdb->query( $wpdb->prepare( "UPDATE $wpdb->posts SET post_title = %s WHERE ID = %d", $var, $id ) );
 ```
 
-`%s` is used for string placeholders and `%d` is used for integer placeholders. Note that they are not 'quoted'! `$wpdb->prepare()` will take care of escaping and quoting for us. The benefit of this is that it is easy to see at a glance whether something has been escaped or not because it happens right when the query happens.
+The following placeholders can be used in the query string:
+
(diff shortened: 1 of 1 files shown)


wordpress-coding-standards/php.md
```diff
@@ -866,7 +866,14 @@ $id = some_foo_number(); // Data we expect to be an integer, but we're not certa
...
 
-`%s` is used for string placeholders and `%d` is used for integer placeholders. Note that they are not 'quoted'! `$wpdb->prepare()` will take care of escaping and quoting for us. The benefit of this is that it is easy to see at a glance whether something has been escaped or not because it happens right when the query happens.
+The following placeholders can be used in the query string:
+
+- %d (integer)
+- %f (float)
+- %s (string)
+- %i (identifier, e.g. table/field names)
+
+`$wpdb->prepare()` will take care of escaping and quoting for us. This makes it easy to see at a glance whether something has been escaped or not because it happens right when the query happens.
```
jrfnl, line 876:
```suggestion
Note that these placeholders should not be 'quoted'! `$wpdb->prepare()` will take care of escaping and quoting for us. This makes it easy to see at a glance whether something has been escaped or not because it happens right when the query happens.
```

I think that warning is important to keep in the manual. I know it is implied in the next phrase, but I'd be rather be very explicit about it, than risk people doing it wrong.

jrfnl (approved):
Thanks @craigfrancis !

dingo-d (approved):

PR #135, opened 2024-02-12
Update php.md

themattyg:
Short echo tags have no reason to be not allowed anymore. WordPress minimum PHP requirement is now 7.4, not <5.6, which seems to be the latest reason to not allow short echo tags. Reference: https://github.com/WordPress/WordPress-Coding-Standards/issues/1642

### DIFF
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -51,20 +51,20 @@ if ( $a === $b ) { ?>
 
 ### No Shorthand PHP Tags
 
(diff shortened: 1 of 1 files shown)


jrfnl:
@themattyg Thanks for your interest in changing the coding standards, however, this is not the way to do it.

Coding standards changes have to be proposed in a Make post and approved, before the handbook can be adjusted.

P.S.: The WordPress minimum PHP requirement is currently 7.0, not 7.4.

PR #136, opened 2024-04-03
Update html.md

joedolson:
The XHTML specification was superseded in 2018, and the HTML specification is now considered the standard of reference. 

As such, while we can still reference the XHTML specification, there are places where XHTML standards conflict with HTML standards, such as the handling of boolean attributes. We should update the HTML standards to explicitly allow boolean attributes. See https://core.trac.wordpress.org/ticket/60178 for reference.

### DIFF
--- a/wordpress-coding-standards/html.md
+++ b/wordpress-coding-standards/html.md
@@ -55,7 +55,19 @@ Incorrect:
 <input type=text name=email disabled>
 ```
 
-In HTML, attributes do not all have to have values, and attribute values do not always have to be quoted. While all of the examples above are valid HTML, _failing to quote attributes can lead to security vulnerabilities_. Always quote attributes.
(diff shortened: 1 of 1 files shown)


dmsnell:
See also Core-59883.

There are multiple places we do awkward things to support content types that WordPress and browsers stopped supporting long ago.

Without HTTP headers or a path ending in `.xml` a browser will interpret _as HTML5_ any document claiming to be XHTML.

aaronjorbin (approved):

aaronjorbin:
LGTM

dingo-d (approved):

wordpress-coding-standards/html.md
```diff
@@ -55,7 +55,19 @@ Incorrect:
...
+Correct:
+
+```html
+<input type="text" name="email" disabled />
+```
+
+Incorrect:
+
+```html
+<input type="text" name="email" disabled="true" />
```
dmsnell, line 69: we may want to also consider that the trailing self-closing flag is also invalid, and remove it from the examples.

```html
<input type="text" name="email" disabled>
```

joedolson: The HTML markup syntax does permit self-closing on void elements; see https://www.w3.org/TR/2011/WD-html-markup-20110113/syntax.html

dmsnell: yes, it is a benign error, like most forms of invalid HTML.

for reference, the living HTML spec is at https://html.spec.whatwg.org

PR #141, opened 2024-09-01
Update accessibility.md

joedolson:
Update the Accessibility Coding Standards and related references to WCAG 2.2. WCAG 2.2 was established as the most recent version of the official W3C recommendations in October 2023, and it is past time that we migrate our standards to coordinate. They are already established in law in the UK, and the European Accessibility Act that comes into enforcement in June 2025 has revisions in progress to update those requirements.

### DIFF
--- a/wordpress-coding-standards/accessibility.md
+++ b/wordpress-coding-standards/accessibility.md
@@ -1,6 +1,6 @@
 # Accessibility Coding Standards
 
-Code integrated into the WordPress ecosystem - including WordPress core, WordPress.org websites, and official plugins, is expected to conform to the Web Content Accessibility Guidelines (WCAG), version 2.1, at level AA.
+Code integrated into the WordPress ecosystem - including WordPress core, WordPress.org websites, and official plugins, is expected to conform to the Web Content Accessibility Guidelines (WCAG), version 2.2, at level AA.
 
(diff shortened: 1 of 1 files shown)


jrfnl:
@joedolson Looks like all that's changed is the version number. Is this correct ?

Does WCAG 2.2 add additional requirements ? If os, should those be included in the docs ? And if so, should this first be posted on Make ?

joedolson:
There are six additional relevant success criteria; reviewable here: https://www.w3.org/WAI/standards-guidelines/wcag/new-in-22/

The existing standards don't document every individual success criteria; only the guidelines, and there were no new guidelines added, so nothing else needs to be added to that document. 

For the most part, we already meet these (at least, as well as we do the existing guidelines), so it's not a major change. One of the more notable issues (3.3.8) is already milestoned for 6.7. 

Yes, we should post about this on Make; and I'm happy to write that up.

joedolson:
It's mostly just the version number; the link to the authoring practices has also changed, and that's no longer directly tied to a version number.

jrfnl:
Thanks for clarifying this @joedolson. A Make post sounds great. Maybe round the 6.7 release ?

jrfnl (approved):

joedolson:
I'll plan on that.

dingo-d (approved):

PR #144, opened 2024-09-10
require/import vs require/include in PHP Coding Standards Documentation. - Fix for #143

### LINKED ISSUE #143: require/import vs require/include in PHP Coding Standards Documentation.
Hello everyone. I stumpled upon this site.

https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#using-import-use-statements

"Note that, unless you have implemented [autoloading](https://www.php.net/manual/en/language.oop5.autoload.php), the use statement won’t automatically load whatever is being imported. You’ll either need to set up autoloading or load the file containing the class/function/constant using a `require/import` statement, for the imported constructs to be loaded when used."

Unless I am mistaken that should read `require/include`?

If `require/import` is correct, just close the issue. I will prepare a PR right away with a fix.

apermo:
This fixes #143 

Also taken care of @jrfnl 's comment on slack that autoloading only applies to classes and not to functions or constants.

### DIFF
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -605,7 +605,8 @@ Group `use` statements are available from PHP 7.0, and trailing commas in group 
 [/alert]
 
 [info]
-Note that, unless you have implemented [autoloading](https://www.php.net/manual/en/language.oop5.autoload.php), the `use` statement won't automatically load referenced classes. You'll either need to set up autoloading or load the file containing the class using `require[_once]` or `include[_once]` statement, for the imported classes to be loaded when used. Autoloading is only applicable to classes; for functions and constants, you must always use `require[_once]` or `include[_once]`.
+Note that, unless you have implemented [autoloading](https://www.php.net/manual/en/language.oop5.autoload.php), the `use` statement won't automatically load whatever is being imported. For OO constructs, you'll either need to set up autoloading or load the file(s) containing the OO declaration(s) using `require[_once]` or `include[_once]` statements.
+Autoloading is only applicable to OO constructs; for functions and constants, you must always use `require[_once]` or `include[_once]`.
 [/info]
 
 **Note about WordPress Core usage**
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -82,7 +82,7 @@ Text that goes into HTML or XML attributes should be escaped so that single or d
 
 Because `require[_once]` and `include[_once]` are language constructs, they do not need parentheses around the path, so those shouldn't be used. There should only be one space between the path and the require/include keywords.
 
-It is _strongly recommended_ to use `require[_once]` for unconditional includes. When using `include[_once]`, PHP will throw a warning when the file is not found but will continue execution, which will almost certainly lead to other errors/warnings/notices being thrown if your application depends on the file loaded, potentially leading to security leaks, or result in silent malfunctions which are hard to track down. For that reason, `require[_once]` is generally the better choice as it will throw a `Fatal Error` if the file cannot be found.
+It is _strongly recommended_ to use `require[_once]` for unconditional includes. When using `include[_once]`, PHP will throw a warning when the file is not found but will continue execution, which will almost certainly lead to other errors/warnings/notices being thrown if your application depends on the file loaded, potentially leading to security leaks. For that reason, `require[_once]` is generally the better choice as it will throw a `Fatal Error` if the file cannot be found.
 
 ```php
 // Correct.
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -605,7 +605,7 @@ Group `use` statements are available from PHP 7.0, and trailing commas in group 
 [/alert]
 
 [info]
-Note that, unless you have implemented [autoloading](https://www.php.net/manual/en/language.oop5.autoload.php), the `use` statement won't automatically load whatever is being imported. You'll either need to set up autoloading or load the file containing the class/function/constant using a `require/include` statement, for the imported constructs to be loaded when used.
+Note that, unless you have implemented [autoloading](https://www.php.net/manual/en/language.oop5.autoload.php), the `use` statement won't automatically load referenced classes. You'll either need to set up autoloading or load the file containing the class using a `require/include` statement, for the imported classes to be loaded when used.Autoloading is only applicable to classes; for functions and constants, you must always use `require` or `include`.
 [/info]
 
 **Note about WordPress Core usage**
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -605,7 +605,7 @@ Group `use` statements are available from PHP 7.0, and trailing commas in group 
 [/alert]
 
 [info]
-Note that, unless you have implemented [autoloading](https://www.php.net/manual/en/language.oop5.autoload.php), the `use` statement won't automatically load whatever is being imported. For OO constructs, you'll either need to set up autoloading or load the file(s) containing the OO declaration(s) using `require[_once]` or `include[_once]` statements.
+Note that, unless you have implemented [autoloading](https://www.php.net/manual/en/language.oop5.autoload.php), the `use` statement won't automatically load whatever is being imported. For OO constructs, you'll either need to set up autoloading or load the file(s) containing the OO declaration(s) using `require[_once]` or `include[_once]` statement(s).
 Autoloading is only applicable to OO constructs; for functions and constants, you must always use `require[_once]` or `include[_once]`.
 [/info]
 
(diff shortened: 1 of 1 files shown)


wordpress-coding-standards/php.md
```diff
@@ -605,7 +605,7 @@ Group `use` statements are available from PHP 7.0, and trailing commas in group
 [/alert]
 
 [info]
-Note that, unless you have implemented [autoloading](https://www.php.net/manual/en/language.oop5.autoload.php), the `use` statement won't automatically load whatever is being imported. You'll either need to set up autoloading or load the file containing the class/function/constant using a `require/import` statement, for the imported constructs to be loaded when used.
+Note that, unless you have implemented [autoloading](https://www.php.net/manual/en/language.oop5.autoload.php), the `use` statement won't automatically load referenced classes. You'll either need to set up autoloading or load the file containing the class using a `require/include` statement, for the imported classes to be loaded when used. Autoloading is only applicable to classes; for functions and constants, you must always use `require` or `include`.
```
jrfnl, line 608:
```suggestion
Note that, unless you have implemented [autoloading](https://www.php.net/manual/en/language.oop5.autoload.php), the `use` statement won't automatically load referenced classes. You'll either need to set up autoloading or load the file containing the class using a `require/include[_once]` statement. Autoloading is only applicable to classes; for functions and constants, you must always use `require[_once]` or `include[_once]`.
```

jrfnl (commented):
Thanks for catching that typo @apermo! I've left a small suggestion inline. Let me know what you think.

wordpress-coding-standards/php.md
apermo: I definitely like the added [_once] in the new sentence. I'll already apply that second part.

I am personally unsure about the `require/include[_once]`.

I personally read it as "'require/include` statement", and due to that "statement" I included the optional `_once` for both. 
If you want to make it clear, I personally would suggest:

[...] or load the file containing the class using `require[_once]` or `include[_once]`. [...]

Imho that improves readability. Otherwise I personally would keep `require/include`.

Your thoughts?

apermo: @jrfnl I've added the before mentioned change. And changed order to alphabetical (first include then require) this matches the paragraph on "Writing include/require statements" in lines 81-83 of the same file.

Let me know your thoughts.

jrfnl: The order was intentional - `require` should generally be preferred over `include`, which is why it was mentioned first.

apermo: @jrfnl I agree with that point. What do you think about adjusting the other appearance of include/require to require/include, to match the same idea?

jrfnl: I'd need to look at the complete doc to be sure (on the road now, so can't look), but in principle, I'd be open to such a change.

apermo: @jrfnl I've [updated the order](https://github.com/WordPress/wpcs-docs/pull/144/commits/e520d9e02f1761b3ade93c5f0753882335fa5775), and I've added a subsentence to the documentation in the upper paragraph. My experience showed that `include` statements can result in silent malfunctions, which are a pain to track down. If you disagree adding this, I'm absolutely fine to omit that.

dingo-d (approved):

wordpress-coding-standards/php.md
```diff
@@ -605,7 +605,7 @@ Group `use` statements are available from PHP 7.0, and trailing commas in group
 [/alert]
 
 [info]
-Note that, unless you have implemented [autoloading](https://www.php.net/manual/en/language.oop5.autoload.php), the `use` statement won't automatically load whatever is being imported. You'll either need to set up autoloading or load the file containing the class/function/constant using a `require/import` statement, for the imported constructs to be loaded when used.
+Note that, unless you have implemented [autoloading](https://www.php.net/manual/en/language.oop5.autoload.php), the `use` statement won't automatically load referenced classes. You'll either need to set up autoloading or load the file containing the class using `require[_once]` or `include[_once]` statement, for the imported classes to be loaded when used. Autoloading is only applicable to classes; for functions and constants, you must always use `require[_once]` or `include[_once]`.
```
jrfnl, line 608: There seems to be some confusion in this sentence between singular and plural - "load referenced classes" versus "the file containing the class".

I think this still needs some finetuning.

Also, while I understand the change from "whatever is being imported" to "referenced classes", it disregards that more can be imported via a `use` statement, including namespaces.

```suggestion
Note that, unless you have implemented [autoloading](https://www.php.net/manual/en/language.oop5.autoload.php), the `use` statement won't automatically load whatever is being imported. For OO constructs, you'll either need to set up autoloading or load the file(s) containing the OO declaration(s) using `require[_once]` or `include[_once]` statements.
Autoloading is only applicable to OO constructs; for functions and constants, you must always use `require[_once]` or `include[_once]`.
```

This may need a check how OO constructs are referred to elsewhere in the document - I'm not sure we make it explicit that this means "classes, interfaces, traits and enums" and/or whether that needs to be repeated.

Also, the above is still confusing if read strictly with a technical mindset, as autoloading === loading the file containing the OO declarations using `require[_once]` or `include[_once]` statements. It is just a way of doing that "just in time"/only when the file is needed.

wordpress-coding-standards/php.md
```diff
@@ -78,11 +78,11 @@ echo "<a href='{$escaped_link}'>text with a ' single quote</a>";
 
 Text that goes into HTML or XML attributes should be escaped so that single or double quotes do not end the attribute value and invalidate the HTML, causing a security issue. See [Data Validation](https://developer.wordpress.org/plugins/security/data-validation/) in the Plugin Handbook for further details.
 
-### Writing include/require statements
+### Writing require/include statements
 
-Because `include[_once]` and `require[_once]` are language constructs, they do not need parentheses around the path, so those shouldn't be used. There should only be one space between the path and the include/require keywords.
+Because `require[_once]` and `include[_once]` are language constructs, they do not need parentheses around the path, so those shouldn't be used. There should only be one space between the path and the require/include keywords.
```
jrfnl, line 83: I don't think the order needs to change here as the paragraph directly below it points out that `require` should be preferred over `include`.

Not a blocker though.

wordpress-coding-standards/php.md
```diff
@@ -78,11 +78,11 @@ echo "<a href='{$escaped_link}'>text with a ' single quote</a>";
...
 
-It is _strongly recommended_ to use `require[_once]` for unconditional includes. When using `include[_once]`, PHP will throw a warning when the file is not found but will continue execution, which will almost certainly lead to other errors/warnings/notices being thrown if your application depends on the file loaded, potentially leading to security leaks. For that reason, `require[_once]` is generally the better choice as it will throw a `Fatal Error` if the file cannot be found.
+It is _strongly recommended_ to use `require[_once]` for unconditional includes. When using `include[_once]`, PHP will throw a warning when the file is not found but will continue execution, which will almost certainly lead to other errors/warnings/notices being thrown if your application depends on the file loaded, potentially leading to security leaks, or result in silent malfunctions which are hard to track down. For that reason, `require[_once]` is generally the better choice as it will throw a `Fatal Error` if the file cannot be found.
```
jrfnl, line 85: I'm not sure the addition adds value in this way and it distracts from/downplays the point about the risk of security leaks.

Errors are already mentioned earlier in the sentence. I'm open to rephrasing that part, but the way it is phrased now, the "For that reason" at the start of the next sentence would predominantly apply to the "silent malfunctions", while it is - predominantly - about the risk of security leaks.

jrfnl (commented):
@apermo Sorry for the slow turn-around. I've had another critical look and I think this still needs more tweaking, largely because the PR is trying to do more than just fix the issue you reported.

wordpress-coding-standards/php.md
apermo: I've changed this for consistency, I agree that it is redundant, but my OCD want to have a consistent order throughout the file. I don't have hard feelings about it either, so if you disagree we can as well omit this change.

wordpress-coding-standards/php.md
apermo: You're right, after reading it again it is kind of redundant with "it will throw a `Fatal Error` if the file cannot be found."

I'll remove it :)

wordpress-coding-standards/php.md
apermo: For consistency I would prefer this, also this will emphasize require over include.
But as mentioned, this is mostly personal preference, so in the end your choice.

apermo:
@jrfnl I've missed your reply and just taken care of your comments. 
Applied the suggestion, removed the "silent error" suggestion.

jrfnl (commented):
Thanks for making those update @apermo ! This is looking good now. Just one nitpick left.

apermo:
Resolved @jrfnl

dingo-d (approved):
Looks good to me 👍🏼

PR #147, opened 2025-03-07
Fix incorrect magic constant __dIr__ to __DIR__

Rat01047:
## Description:
This PR fixes a typo in the magic constant __dIr__, which is invalid.

### Fixes:
-> The correct constant __DIR__ is now used to ensure proper path resolution.

### DIFF
--- a/wordpress-coding-standards/php.md
+++ b/wordpress-coding-standards/php.md
@@ -464,7 +464,7 @@ add_action( 'action_name', array( __CLASS__, 'method_name' ) );
 add_action( 'action_name', array( My_Class::class, 'method_name' ) );
 
 // Incorrect.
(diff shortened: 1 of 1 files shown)


dingo-d (approved):
Thanks for the typo fix, not sure how this slipped through 🙈

jrfnl:
Not a typo. It is one of the "Incorrect" examples for the following rule:

> The [PHP native `__*__` magic constants](https://www.php.net/manual/en/language.constants.magic.php), like `__CLASS__` and `__DIR__`, should be written in uppercase when used.

Oh and in case you are wondering: PHP handles this without any problems, so there is no issue with "proper path resolution"....

Closing as invalid.

PR #148, opened 2025-05-06
Update php.md

mehrazmorshed:


dingo-d (approved):

GaryJones (approved):

jrfnl:
Why was this merged ? There was nothing wrong with the code sample as it was...

dingo-d:
> Why was this merged ? There was nothing wrong with the code sample as it was...

Technically, no, but I think it's clearer to see the error if the case is the same. Easier to spot it imo.

jrfnl:
@dingo-d Fair enough. The example was deliberately like it was to show a typical example of people aligning the colons to the longest case name as incorrect. That's now gone.

dingo-d:
> The example was deliberately like it was to show a typical example of people aligning the colons to the longest case name as incorrect.

Wasn't aware of that intention.

> That's now gone.

We can always add a similar example in a new PR.

jrfnl:
> We can always add a similar example in a new PR.

I rather think this commit should be reverted. A `switch` with two `case`s for the same is not a very realistic code sample anyway, so this should never have been merged.

dingo-d:
> I rather think this commit should be reverted. A switch with two cases for the same is not a very realistic code sample anyway, so this should never have been merged.

I looked at the examples in a way of 'right' and 'wrong' way on the same piece of code, not as a one functional code example 🤷🏼‍♂️

PR #150, opened 2025-06-01
Use variable names of different lengths in DocBlock example

sabernhardt:
If one global variable's name is longer than the other in the group, that can help demonstrate how they are "aligned by type, variable, and description."

- My suggestions for the names could be better.
- The `param` examples might benefit from a similar revision.

### DIFF
--- a/inline-documentation-standards/php.md
+++ b/inline-documentation-standards/php.md
@@ -206,8 +206,8 @@ Functions and class methods should be formatted as follows:
  *
  * @see Function/method/class relied on
  * @link URL
- * @global type $varname Description.
- * @global type $varname Description.
+ * @global type $variable_name    Description.
+ * @global type $another_variable Description.
  *
  * @param type $var Description.
  * @param type $var Optional. Description. Default.


jrfnl (commented):
Should the same be done for the `@param` tags directly below ?

sabernhardt:
`$wpdb` is one of the most common global variables, and `switch_to_blog()` has other globals—of various types—to consider for the example:
```
 * @global wpdb   $wpdb         WordPress database abstraction object.
 * @global int    $blog_id
 * @global bool   $switched
 * @global string $table_prefix The database table prefix.
```

Then the parameters could use something like `$variable_name` and `$another_variable`:
```
 * @param type $variable_name    Description.
 * @param type $another_variable Optional. Description. Default.
 * @return type Description.
```

jrfnl:
@sabernhardt Thank you for coming up with those examples.

Seeing the examples you posted above.... I realize my remark about using existing WP globals was wrong, as it would lead to the example having an actual description, instead of just "Description".
I believe this will distract from what the code sample is trying to demonstrate (alignment), so the `@global` example should probably stay as you already currently have it in this PR.

As for the variable names, I can live with the current ones. We can also change it to more typical "anonymous" examples like `$foo` and `$barbar`. What do you think ?

For the parameter example, the names could possible be changed to `$required_param` and `$optional` ? - still not pretty, but the terms "required" and "optional" are unfortunately the same length, so just `$required` and `$optional` would still not get us different sized variables.

PR #159, opened 2026-05-03
Introduce the GitHub Actions Workflow Standards page

johnbillion:
Work has been ongoing recently to improve the quality and security of GitHub Actions workflow files in the `wordpress-develop` and `gutenberg` repos. Examples:

- https://github.com/WordPress/gutenberg/pull/69126
- https://github.com/WordPress/wordpress-develop/pull/8007
- https://github.com/WordPress/wordpress-develop/pull/9767

While the Actionlint and Zizmor tools that we now use are great for surfacing problems in workflow files, we also need good user-facing documentation that provides guidance on addressing common issues and what automation is in place.

Members of the security team have been working on this page and it's already live at https://developer.wordpress.org/coding-standards/wordpress-coding-standards/github-actions/. It now needs to be moved into the docs repo alongside the other pages.

### DIFF
--- a/wordpress-coding-standards.md
+++ b/wordpress-coding-standards.md
@@ -13,6 +13,7 @@ If you are planning to contribute to WordPress core, you need to familiarize you
 <h2>Language-specific Standards</h2>
 <ul>
 		<li><a href="https://developer.wordpress.org/coding-standards/wordpress-coding-standards/css/">CSS Coding Standards</a></li>
+		<li><a href="https://developer.wordpress.org/coding-standards/wordpress-coding-standards/github-actions/">GitHub Actions Workflow Standards</a></li>
 		<li><a href="https://developer.wordpress.org/coding-standards/wordpress-coding-standards/html/">HTML Coding Standards</a></li>
 		<li><a href="https://developer.wordpress.org/coding-standards/wordpress-coding-standards/javascript/">JavaScript Coding Standards</a></li>
 		<li><a href="https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/">PHP Coding Standards</a></li>
--- a/wordpress-coding-standards.md
+++ b/wordpress-coding-standards.md
@@ -13,7 +13,6 @@ If you are planning to contribute to WordPress core, you need to familiarize you
 <h2>Language-specific Standards</h2>
 <ul>
 		<li><a href="https://developer.wordpress.org/coding-standards/wordpress-coding-standards/css/">CSS Coding Standards</a></li>
-		<li><a href="https://developer.wordpress.org/coding-standards/wordpress-coding-standards/github-actions/">GitHub Actions Workflow Standards</a></li>
 		<li><a href="https://developer.wordpress.org/coding-standards/wordpress-coding-standards/html/">HTML Coding Standards</a></li>
 		<li><a href="https://developer.wordpress.org/coding-standards/wordpress-coding-standards/javascript/">JavaScript Coding Standards</a></li>
 		<li><a href="https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/">PHP Coding Standards</a></li>
@@ -26,3 +25,9 @@ WordPress is committed to meeting the <a href="https://www.w3.org/TR/WCAG20/">We
 <h2>Where do the coding standards _not_ apply?</h2>
 
 Third-party libraries are not subject to these standards, even when integrated with the primary project. This includes instances like WordPress core, where multiple third-party libraries are incorporated into its codebase.
+
+<h2>GitHub Actions Workflow Standards</h2>
+
+GitHub Actions workflows operate in a highly privileged software supply chain environment.
+
+All repos under the WordPress organisation on GitHub are expected to adhere to <a href="https://developer.wordpress.org/coding-standards/wordpress-coding-standards/github-actions/">the GitHub Actions Workflow Standards</a>.
--- a/wordpress-coding-standards/github-actions.md
+++ b/wordpress-coding-standards/github-actions.md
@@ -159,7 +159,8 @@ If the job needs persistent credentials (for example, to push built files), set 
 
 Writing to `$GITHUB_ENV` or `$GITHUB_PATH` from a shell script is dangerous if the input is user-controlled, because an attacker can inject arbitrary environment variables or prepend to `PATH`.
 
-- Only write to `$GITHUB_ENV` or `$GITHUB_OUTPUT` with values that are fully controlled by the workflow, not with values derived from pull request content, issue bodies, commit messages, or other user-controllable inputs.
+- To pass state between steps, prefer writing to `$GITHUB_OUTPUT` instead. Unlike `$GITHUB_ENV` and `$GITHUB_PATH`, outputs are not injected into the environment or `PATH` of subsequent steps.
+- If you must write to `$GITHUB_ENV` or `$GITHUB_PATH`, only write values that are fully controlled by the workflow, not with values derived from pull request content, issue bodies, commit messages, or other user-controllable inputs.
 - If you must process user-controllable input, validate and sanitize it before writing to these files.
 
 ### Unpinned uses
--- a/wordpress-coding-standards/github-actions.md
+++ b/wordpress-coding-standards/github-actions.md
@@ -1,0 +1,185 @@
+# GitHub Actions Workflow Standards
+
+GitHub Actions workflows operate in a highly privileged software supply chain environment. Workflows can access repository secrets, push code, create releases, publish packages, and interact with external services. A security weakness in a workflow file can have severe consequences.
+
+WordPress uses two complementary linting tools to help maintain the quality and security of workflow files in the `.github/workflows` directory: [Actionlint](https://github.com/rhysd/actionlint) and [Zizmor](https://github.com/zizmorcore/zizmor). This page documents the tools and how contributors should address errors or warnings that they report.
+
+## Actionlint
+
+Actionlint is a static checker for workflow files. It focuses primarily on correctness: syntax validation, type checking for expressions, validation of inputs for actions and reusable workflows, syntax checking of shell scripts, and other common mistakes. See the [Actionlint documentation](https://github.com/rhysd/actionlint) for details.
+
+Actionlint runs on pull requests and on pushes to the main branches on [the wordpress-develop repo](https://github.com/WordPress/wordpress-develop). It reports its findings as check results, just like the unit test and coding standards workflows. A failing Actionlint check must be fixed before the changes in the PR can be committed.
+
+## Zizmor
+
+Zizmor is a security-focused linter for workflow files. It detects template injection vulnerabilities, excessive permissions, dangerous triggers, unpinned dependencies, credential persistence, and dozens of other security weaknesses. See the [Zizmor documentation](https://docs.zizmor.sh/audits/) for details.
+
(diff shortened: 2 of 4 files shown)


wordpress-coding-standards/github-actions.md
```diff
@@ -0,0 +1,185 @@
...
+From the root of the repository, run (note the trailing period):
+
+```
+zizmor .
+```
+
+To enable the online audits that check for known-vulnerable actions and impostor commits, provide a GitHub token:
+
+```
+GH_TOKEN=$(gh auth token) zizmor .
```
dingo-d: Maybe a note should be added here that [GitHub CLI](https://cli.github.com/) should be installed for this command to work

dingo-d (commented):
I had just one comment, otherwise looks good 👍🏼

wordpress-coding-standards/github-actions.md
johnbillion: Done!

dingo-d (approved):

jrfnl:
@johnbillion You seem to be mixing two different tasks in this single PR:
* Adding a new page
* Improving the security of the GH Actions workflow(s) for this repo.

As those are different decision points, with different justifications, could you please move the workflow changes to their own PR ?

johnbillion:
> could you please move the workflow changes to their own PR ?

Done: https://github.com/WordPress/wpcs-docs/pull/161

wordpress-coding-standards/github-actions.md
```diff
@@ -0,0 +1,185 @@
...
+```yaml
+- name: Print title
+  run: echo "Title: ${{ github.event.pull_request.title }}"
+```
+
+**Correct:**
+
+```yaml
+- name: Print title
+  run: echo "Title: ${PR_TITLE}"
```
desrosj: Small nit and admittedly a personal preference, but I prefer that `run:` comes last in the step definition because it ensures the reader notices all conditions, variables, environments, etc. before reading the actual command that is run.

wordpress-coding-standards/github-actions.md
```diff
@@ -0,0 +1,185 @@
...
+- uses: actions/github-script@...
+  with:
+    script: |
+      const title = "${{ github.event.pull_request.title }}";
+```
+
+**Correct:**
+
+```yaml
+- uses: actions/github-script@...
```
desrosj: Same preference comment here about the `uses:` being last. However, I feel more strongly that any inputs passed should be immediately after the `uses:` line.

wordpress-coding-standards/github-actions.md
```diff
@@ -0,0 +1,185 @@
...
+    persist-credentials: false
+```
+
+If the job needs persistent credentials (for example, to push built files), set `persist-credentials: true` explicitly so the intent is clear and auditable, and include an accompanying comment.
+
+### GitHub environment manipulation
+
+Writing to `$GITHUB_ENV` or `$GITHUB_PATH` from a shell script is dangerous if the input is user-controlled, because an attacker can inject arbitrary environment variables or prepend to `PATH`.
+
+- Only write to `$GITHUB_ENV` or `$GITHUB_OUTPUT` with values that are fully controlled by the workflow, not with values derived from pull request content, issue bodies, commit messages, or other user-controllable inputs.
```
desrosj: Should we include an example about how to do this properly by wrapping `$GITHUB_OUTPUT` in double quotes?

wordpress-coding-standards/github-actions.md
```diff
@@ -0,0 +1,185 @@
...
+- uses: actions/checkout@v6
+```
+
+**Correct:**
+
+```yaml
+- uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+```
+
+Always include a version comment after the SHA to make the pinned version human-readable. When updating an action, update both the SHA and the version comment.
```
desrosj: May be worth noting that the version specified within the online comment must match the actual version tag published by the action's repository.

wordpress-coding-standards/github-actions.md
```diff
@@ -0,0 +1,185 @@
...
+
+### Cache poisoning
+
+Using GitHub Actions caching in workflows that produce release artifacts is risky. A cache can be poisoned by an attacker in a separate workflow, allowing the poisoned cache to inject malicious content into a release.
+
+Avoid using `actions/cache` or built-in caching features in workflows that build and publish packages or release artifacts. If caching is necessary in such workflows, ensure the cache key is scoped tightly and the cache contents are verified before use.
```
desrosj: Cache poisoning can also occur when a workflow specifies `restore-keys`. We should explicitly discourage the use of partial cache key matching patterns in favor of explicit hash-based keys where the hash was produced using the relevant lock files.

desrosj (commented):
@johnbillion some small suggestions. Some may be a bit deeper than you're looking to go in this high-level guide, and some are stylistic preferences.

wordpress-coding-standards/github-actions.md
rodrigoprimo: The paragraph above flags writing to `$GITHUB_ENV` or `$GITHUB_PATH`, but this line mentions writing to `$GITHUB_ENV` or `$GITHUB_OUTPUT`. Should it say `$GITHUB_PATH` instead of `$GITHUB_OUTPUT`? I might be missing something, as I'm not super familiar with how GitHub Actions uses those environment variables.

wordpress-coding-standards/github-actions.md
```diff
@@ -0,0 +1,185 @@
...
+
+For full information about specific results from the linting tools, consult [the Actionlint documentation](https://github.com/rhysd/actionlint/blob/main/docs/checks.md) and [the Zizmor documentation](https://docs.zizmor.sh/audits/).
+
+### Template injection
+
+Template injection occurs when a GitHub Actions expression such as `${{ github.event.issue.title }}` is used directly within a `run:` block. GitHub Actions expressions are interpreted _prior_ to running the script, therefore an attacker who controls the expression value can inject arbitrary shell commands, regardless of whether the expression is wrapped in quotes.
+
+**Incorrect:**
+
+```yaml
```
rodrigoprimo, line 66: The YAML code blocks on the [published page](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/github-actions/) are not syntax-highlighted. They render as `<code class="">` rather than `<code class="language-yaml">` like the PHP and CSS blocks do on other pages. Do you know if ```yaml fences work with the importer? I don't know whether the published page was generated from a version of this Markdown file.

rodrigoprimo (commented):
Looks good to me overall. I left a couple of inline comments with questions about small things that caught my attention.

wordpress-coding-standards.md
```diff
@@ -13,6 +13,7 @@ If you are planning to contribute to WordPress core, you need to familiarize you
 <h2>Language-specific Standards</h2>
 <ul>
 		<li><a href="https://developer.wordpress.org/coding-standards/wordpress-coding-standards/css/">CSS Coding Standards</a></li>
+		<li><a href="https://developer.wordpress.org/coding-standards/wordpress-coding-standards/github-actions/">GitHub Actions Workflow Standards</a></li>
```
GaryJones: It’s slotted under “Language-specific Standards” alongside CSS/HTML/JS/PHP. GitHub Actions isn’t a language; it’s YAML config for a CI platform. This will read oddly to anyone scanning the index, and it implies a parity (“this is a WordPress coding standard”) that the content doesn’t quite live up to. A separate “Infrastructure / Tooling Standards” grouping would be more honest, and would leave room for similar future docs (Composer, npm, Docker).

wordpress-coding-standards/github-actions.md
```diff
@@ -0,0 +1,185 @@
+# GitHub Actions Workflow Standards
+
+GitHub Actions workflows operate in a highly privileged software supply chain environment. Workflows can access repository secrets, push code, create releases, publish packages, and interact with external services. A security weakness in a workflow file can have severe consequences.
+
+WordPress uses two complementary linting tools to help maintain the quality and security of workflow files in the `.github/workflows` directory: [Actionlint](https://github.com/rhysd/actionlint) and [Zizmor](https://github.com/zizmorcore/zizmor). This page documents the tools and how contributors should address errors or warnings that they report.
```
GaryJones, line 5: The scope is undefined. The doc says “WordPress uses…” then only names wordpress-develop as an enforcement point. Does this apply to Gutenberg? Plugin/theme contributors? The wpcs-docs repo itself? The PR body hints at all of them, but the page should state its scope explicitly — otherwise it’s a guideline dressed as a standard.

wordpress-coding-standards/github-actions.md
```diff
@@ -0,0 +1,185 @@
+# GitHub Actions Workflow Standards
```
GaryJones, line 1: There's a title vs content imbalance here. It's called “GitHub Actions Workflow Standards”, but the substance is almost entirely Zizmor audit findings rephrased. Actionlint gets one paragraph and zero concrete examples. Either rename it to something like “Workflow Security Standards”, or beef up the Actionlint side (common findings, shellcheck guidance, expression typing pitfalls).

For a document presented as the standard, I’d expect at least a mention of:

- timeout-minutes (DoS / runaway jobs)
- Concurrency controls (`concurrency:` block)
- GITHUB_TOKEN scoping vs PATs
- etc.

wordpress-coding-standards/github-actions.md
```diff
@@ -0,0 +1,185 @@
...
+permissions: {}
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+```
+
+### Artipacked credentials
```
GaryJones: This is a Zizmor audit name. For readers who’ve never used Zizmor, it’s jargon. “Persistent checkout credentials (Zizmor: artipacked)” reads more naturally.

wordpress-coding-standards/github-actions.md
GaryJones: “Scope the cache key tightly and verify the cache contents before use” — how? A concrete example or a hard rule (“don’t cache in release workflows, full stop”) would be more useful than the current half-advice.

GaryJones (dismissed):
Also, the PR body still says "This PR also fixes some issues in the qa.yml workflow file in this repo so it adheres to these standards.", even though that was moved out.

wordpress-coding-standards/github-actions.md
desrosj: The end goal is to add enforcement of these practices org-wide to ensure all repositories remain secure. But that will take time.

That said, we could maybe add a section linking to the repositories enforcing this to date. Gutenberg runs a workflow, but I don't believe it is configured as a check that is required to pass currently. I could be wrong on that and would need to check.

wordpress-coding-standards/github-actions.md
johnbillion: Double quotes protect against word splitting be present in the value of the `GITHUB_OUTPUT` variable itself, they don't protect against user-controlled input being written to the file that `$GITHUB_OUTPUT` normally points to.

We need to balance making this document useful and approachable for the majority of contributors, versus covering every protection that Actionlint and Zizmor affords us. I would prefer to leave out mention of double quoting these values for now.

wordpress-coding-standards/github-actions.md
johnbillion: I updated the code blocks on the published page to use YAML syntax highlighting. The current page was created by pasting in the markdown into the block editor.

wordpress-coding-standards/github-actions.md
johnbillion: Yep all fair points.

This document currently focuses on the security aspect of workflows, but in the future I fully expect it to be extended to include timeout-minutes, concurrency, use of reusable workflows, and maybe even use of `npx` and other software installation and dependency management concerns, but they're all out of scope for this first version.

The current objective is to provide a point of reference for workflow security hardening that repo maintainers can be pointed to.

PR #67, opened 2022-05-02
Remove Changelog page

jrfnl:
As discussed in PR #65, the changelog page is not actively maintained and is out of date/out of sync with the actual changes which have been made, so we may as well remove it.

PR #162, opened 2026-08-17
Update CSS coding standards to adapt to the Stylelint configuration.

afercia:
In https://core.trac.wordpress.org/ticket/29792 / https://github.com/WordPress/wordpress-develop/pull/12934 I'm proposing to add Stylelint configuration and scripts to Core to enforce the CSS coding standards.

The work there is almost done. However, some points of the current CSS coding standards either can't be enforced via Stylelint of hare too difficult to enforce.

As such, I'm proposing a few changes. I will also discuss this with the involved Core teams and consider to publish a post on Make, if necessary.

The most important change is about the class selector pattern. I'm proposing to use the BEM-based naming convention because:
- It is already used in Gutenberg (even though it's based on the component name there).
- It has already slipped through into core, where there are already BEM-like selectors in the about.css and in the wp-tooltip.css files.
- The current naming convention (only hyphens to separate words) is largely not applied across the admin CSS. There are literally hundreds of cases where selectors contain single or double underscores and uppercase characters. They can't be fixed because of backwards compatibility issues. By using the BEM-like new naming convention the numberr of reported eerrors would be reduced. All other cases will need to get a `stylelint-disable` comment.

Other changes are very minor, for example standardize the required number of empty lines to 1. Stylelint isn't abel to apply this rule conditionally to section comments which currently require 2 empty lines. After all, having 2 lines before a section comment doesn't add much value and in my opinion it's best to have a tool that can apply reliably 1 empty line ratehr than requiring exceptions that can't be linted.

### DIFF
--- a/wordpress-coding-standards/css.md
+++ b/wordpress-coding-standards/css.md
@@ -165,7 +165,7 @@ Above all else, choose something that is meaningful to you and semantic in some 
 - Colors and Typography
 - Other
 
-Things that are not yet used in core itself, such as CSS3 animations, may not have a prescribed place above but likely would fit into one of the above in a logical manner. Just as CSS is evolving, so our standards will evolve with it.
+Things that are not yet used in core itself may not have a prescribed place above but likely would fit into one of the above in a logical manner. Just as CSS is evolving, so our standards will evolve with it.
 
 Top/Right/Bottom/Left (TRBL/trouble) should be the order for any relevant properties (e.g. `margin`), much as the order goes in values. Corner specifiers (e.g. `border-radius-*-*`) should be ordered as top-left, top-right, bottom-right, bottom-left. This is derived from how shorthand values would be ordered.
 
@@ -281,7 +281,6 @@ Incorrect:
 Media queries allow us to gracefully degrade the DOM for different screen sizes. If you are adding any, be sure to test above and below the break-point you are targeting.
 
 - It is generally advisable to keep media queries grouped by media at the bottom of the stylesheet.
-    - An exception is made for the `wp-admin.css` file in core, as it is very large and each section essentially represents a stylesheet of its own. Media queries are therefore added at the bottom of sections as applicable.
 - Rule sets for media queries should be indented one level in.
 
 Example:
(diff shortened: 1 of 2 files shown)


jrfnl:
Hi @afercia Thanks for this PR.

This will need a Make post as it changes the guidelines and something like the 1 vs 2 blank lines change may seem like a small change, but it will cause code churn in the current codebase, as well as invalidate patches/cause extra work for committers to update pre-existing patches before they can be committed.

Also, no need to add a changelog entry. The changelog is not used and there is a proposal open to remove it. See #67.
The git history should suffice.

afercia:
Hi @jrfnl 
yes, a Make post was in the plans.
I will remove the changelog entry.

afercia:
A note about a small mistake in the current docs:

`input[type="text"]`

is used as example twice. In [the first example snippet](https://github.com/WordPress/wpcs-docs/blob/a9ffaa51e7c8d3bdf8732deeaea8f38e2d18ab15/wordpress-coding-standards/css.md#L53-L55) is indicated as 'correct'.

while in [the second example snippet](https://github.com/WordPress/wpcs-docs/blob/a9ffaa51e7c8d3bdf8732deeaea8f38e2d18ab15/wordpress-coding-standards/css.md#L77-L79) is indicated (rightly) as 'incorrect' with the explicit explanation: `Should be [type="text"]`.

jrfnl:
> A note about a small mistake in the current docs:
> 
> `input[type="text"]`
> 
> is used as example twice. In [the first example snippet](https://github.com/WordPress/wpcs-docs/blob/a9ffaa51e7c8d3bdf8732deeaea8f38e2d18ab15/wordpress-coding-standards/css.md#L53-L55) is indicated as 'correct'.
> 
> while in [the second example snippet](https://github.com/WordPress/wpcs-docs/blob/a9ffaa51e7c8d3bdf8732deeaea8f38e2d18ab15/wordpress-coding-standards/css.md#L77-L79) is indicated (rightly) as 'incorrect' with the explicit explanation: `Should be [type="text"]`.

I don't think that change is correct. The way I read it, the correct versus incorrect is about the word "text" being between double quotes (correct) or without quotes (incorrect). It has nothing to do with the `input` which you have now removed.

AFAICS, this is an example to demonstrate the following rule listed above it:
> * Attribute selectors should use double quotes around values.

afercia:
> I don't think that change is correct. The way I read it, the correct versus incorrect is about the word "text" being between double quotes (correct) or without quotes (incorrect). It has nothing to do with the `input` which you have now removed.

Yes, it's in an example for _another_ rule. Still, it's very confusing. The same selector is used in a 'correct' example and then in an 'incorrect' example. I'm not sure this helps clarity for users.
At the very least, there should be a comment like the one used in similar cases: `Also doubly incorrect`.

That said, right now there are 1815 violations of 'over-qualified selectors' in Core so... 🤷🏻

wordpress-coding-standards/css.md
```diff
@@ -123,7 +165,7 @@ Above all else, choose something that is meaningful to you and semantic in some
 - Colors and Typography
 - Other
 
-Things that are not yet used in core itself, such as CSS3 animations, may not have a prescribed place above but likely would fit into one of the above in a logical manner. Just as CSS is evolving, so our standards will evolve with it.
+Things that are not yet used in core itself may not have a prescribed place above but likely would fit into one of the above in a logical manner. Just as CSS is evolving, so our standards will evolve with it.
```
afercia, line 170: Removed `such as CSS3 animations,` because they are now used in Core.

wordpress-coding-standards/css.md
```diff
@@ -239,7 +281,6 @@ Incorrect:
 Media queries allow us to gracefully degrade the DOM for different screen sizes. If you are adding any, be sure to test above and below the break-point you are targeting.
 
 - It is generally advisable to keep media queries grouped by media at the bottom of the stylesheet.
-    - An exception is made for the `wp-admin.css` file in core, as it is very large and each section essentially represents a stylesheet of its own. Media queries are therefore added at the bottom of sections as applicable.
 - Rule sets for media queries should be indented one level in.
```
afercia, line 286: Removed all this point because `wp-admin.css` is no longer a huge file split into sections so that the reference to this as an exception doesn't make sense any longer.

afercia:
> This will need a Make post 

Hi @jrfnl 
I prepared a draft on Make for the proposal to update the CSS coding standards. Can I ping you directly on Slack to share it for review?

jrfnl:
> > This will need a Make post
> 
> Hi @jrfnl I prepared a draft on Make for the proposal to update the CSS coding standards. Can I ping you directly on Slack to share it for review?

@afercia I appreciate the ask, but I currently don't have the bandwidth to review such a post. Maybe @GaryJones or @ntwb can help ?

GaryJones:
@afercia Sure!

PR #163, opened 2026-09-04
Document runnable code examples in DocBlocks

adamziel:
Adds runnable PHP example guidance directly to the PHP Documentation Standards page. It documents the `php interactive` fence, `// Outputs` formats, and setup Blueprints used by the Code Reference.

## Background

The PHP documentation standards already explain indented examples, but do not describe the opt-in syntax for examples that readers can run in WordPress Playground. This belongs with the PHP DocBlock rules, where authors already look for example guidance.

## This change

- Adds a `Runnable examples` subsection to `inline-documentation-standards/php.md`, directly after `Description`.
- Documents the fence rules, code requirements, output forms, legacy-fence guidance, and hidden setup examples.
- Keeps the feature opt-in, so existing indented examples continue to work unchanged.
- Does not add a separate handbook page or sidebar item.

## Testing

- Ran `git diff --check`.
- Parsed `manifest.json` and checked the PHP page placement, links, and Markdown fence pairs with a small validation script.


sirreal:
https://make.wordpress.org/core/2026/09/04/runnable-code-examples-are-now-live-in-the-code-reference/

inline-documentation-standards/php.md
```diff
@@ -133,6 +133,115 @@ HTML markup should never be used outside of code examples, though Markdown can b
...
+
+##### The code
+
+Start with the first meaningful line. Do not include `<?php` (the docs site adds it; a snippet starting with it will not run) or `require '/wordpress/wp-load.php';` (WordPress is already loaded).
+
+Print something with `echo`, `print_r()`, or `var_dump()`. Keep it short enough to read comfortably in the source; if it needs a lot of setup, it is a tutorial, not a reference example.
+
+##### The output
+
+End with a `// Outputs` comment. The parser strips it from the code that runs, shows it as the expected result before the reader selects **Run**, and keeps it for later output checks.
```
dingo-d, line 175: I know it's _a comment_, but reading the sentence "End with a outputs comment" just sounds off to me. It sounds better with "End with **an** `// Outputs` comment.".

@GaryJones you're a native speaker, you know best 😄

inline-documentation-standards/php.md
```diff
@@ -133,6 +133,115 @@ HTML markup should never be used outside of code examples, though Markdown can b
...
+
+One line — everything after the colon is literal:
+```php
+echo esc_html( '<egg>' );
+// Outputs: <egg>
+```
```
dingo-d, line 181: Should this be wrapped in four ticks and then inside it have an example of how to write the documentation with `interactive` keyword in it or is this intentional?

inline-documentation-standards/php.md
```diff
@@ -133,6 +133,115 @@ HTML markup should never be used outside of code examples, though Markdown can b
...
+
+Several lines — empty `// Outputs:` starts a block, each `//` line is one output line, a final empty `//` keeps a trailing newline:
+```php
```
dingo-d, line 184: Same here as the above comment.

inline-documentation-standards/php.md
```diff
@@ -133,6 +133,115 @@ HTML markup should never be used outside of code examples, though Markdown can b
...
+//
+```
+
+JSON-encoded — for trailing spaces, tabs, or repeated newlines. Must be one JSON string:
+```php
+echo "done ";
+// Outputs (JSON-encoded): "done "
+```
+
+Only a trailing standalone line comment counts. `// Outputs:` inside a string, a block comment, after code on the same line, or followed by more code is treated as code. The older ` ```expected-output ` fence still parses; do not use both.
```
dingo-d, line 200:
```suggestion
Only a trailing standalone line comment counts. `// Outputs:` inside a string, a block comment, after code on the same line, or followed by more code is treated as code. The older ` ```expected-output ` fence still parses. Do not use both.
```

inline-documentation-standards/php.md
```diff
@@ -133,6 +133,115 @@ HTML markup should never be used outside of code examples, though Markdown can b
...
+
+- `php interactive`, lowercase, in that order. Anything else (`PHP interactive`, `php-interactive`, `php title="x" interactive`) renders as plain code.
+- May be indented, including inside a list item. Code lines need not match the fence indentation.
+- Rendered in place, between the surrounding paragraphs.
+
+##### The code
+
+Start with the first meaningful line. Do not include `<?php` (the docs site adds it; a snippet starting with it will not run) or `require '/wordpress/wp-load.php';` (WordPress is already loaded).
+
+Print something with `echo`, `print_r()`, or `var_dump()`. Keep it short enough to read comfortably in the source; if it needs a lot of setup, it is a tutorial, not a reference example.
```
dingo-d, line 171:
```suggestion
Print something with `echo`, `print_r()`, or `var_dump()`. Keep it short enough to read comfortably in the source. If it needs a lot of setup, it is a tutorial, not a reference example.
```

We are not writing prose, semicolon in docs looks off. Probably an AI artefact.

dingo-d (changes requested):
Left a couple of comments.

juanmaguitar:
I reviewed the PR and think it clearly explains how and when to use runnable examples. 
It’s good to go IMO.

jrfnl:
For some reason I can't seem to add @zzap as a reviewer, but Milana: please consider this an invite to review this PR.

inline-documentation-standards/php.md
```diff
@@ -133,6 +133,115 @@ HTML markup should never be used outside of code examples, though Markdown can b
      * @link https://core.trac.wordpress.org/ticket/20000
     ```
 
+#### Runnable examples
+
+A DocBlock example can be marked as runnable. The Code Reference then shows it with a **Run** button, and readers execute and edit it in the browser through [WordPress Playground](https://developer.wordpress.org/playground/handbook/guides/php-code-snippets/).
+
+This is opt-in. Existing indented examples keep working. Convert an example only when running it teaches something reading it does not.
```
jrfnl, line 140:
```suggestion
This is opt-in. Existing indented examples keep working. Convert an example only, when running it, teaches something reading it does not.
```

inline-documentation-standards/php.md
```diff
@@ -133,6 +133,115 @@ HTML markup should never be used outside of code examples, though Markdown can b
...
+ * // Outputs: free <egg> lang-en
+ * ```
+ */
+````
+
+Three parts: a fenced block, the info string `php interactive`, and a trailing `// Outputs:` comment.
+
+##### The fence
+
+- `php interactive`, lowercase, in that order. Anything else (`PHP interactive`, `php-interactive`, `php title="x" interactive`) renders as plain code.
```
jrfnl, line 163: Is this an official markdown standard or is this a WP markdown specific thing ? If it is an official standard, I believe a link to the specs would be a good addition.

I would ask whether this `interactive` tag addition is supported by markdown parsers and linters, but as we're talking generated docs which are not maintained as markdown (but maintained in the PHP source code), I don't think we need to worry about that.

inline-documentation-standards/php.md
```diff
@@ -133,6 +133,115 @@ HTML markup should never be used outside of code examples, though Markdown can b
...
+- May be indented, including inside a list item. Code lines need not match the fence indentation.
+- Rendered in place, between the surrounding paragraphs.
+
+##### The code
+
+Start with the first meaningful line. Do not include `<?php` (the docs site adds it; a snippet starting with it will not run) or `require '/wordpress/wp-load.php';` (WordPress is already loaded).
```
jrfnl, line 169: Not sure, but wondering if this would be more correct ?

```suggestion
Start with the first meaningful line. Do not include `<?php` (the docs generation script adds it; a snippet starting with it will not run) or `require '/wordpress/wp-load.php';` (WordPress is already loaded).
```

inline-documentation-standards/php.md
jrfnl, line 171:
```suggestion
Print something with `echo`, `print_r()`, `var_export()` or `var_dump()`. Keep it short enough to read comfortably in the source; if it needs a lot of setup, it is a tutorial, not a reference example.
```

jrfnl: Actually, changing this to a separate sentence makes the phrasing unclear.

How about this ?
```suggestion
Print something with `echo`, `print_r()`, or `var_dump()`. Keep it short enough to read comfortably in the source.
Keep in mind, this should should only be used for reference examples. It should not be a tutorial.
```

inline-documentation-standards/php.md
jrfnl, line 175: > keeps it for later output checks

What does this mean ? What does this do ?

Just reading the phrase as it is now, it feels superfluous to add `// Outputs...` as the code is supposed to be interactive and having to declare the expected output feels like it _fakes interaction_ instead of being interactive, as if it were truly interactive, the expected output would not be needed.
I think the explanation of why the `// Outputs ...` comment is needed, and what its function is, needs to be improved.

As an aside, this feels like a maintenance pitfall, but I'm sure you all have considered that.

inline-documentation-standards/php.md
jrfnl: Was just about to ask the same....

inline-documentation-standards/php.md
```diff
@@ -133,6 +133,115 @@ HTML markup should never be used outside of code examples, though Markdown can b
...
+// Outputs:
+// Array
+// (
+//     [fruit] => apple
+// )
+//
+```
+
+JSON-encoded — for trailing spaces, tabs, or repeated newlines. Must be one JSON string:
+```php
```
jrfnl, line 195: Here is another one of the "this is not a docblock, so not representative of what this documentation is trying to explain"... (and there are still more below)

inline-documentation-standards/php.md
jrfnl, line 200: Is this ` ```expected-output ` fence documented anywhere ? If so, where ? If not, should it be ?

inline-documentation-standards/php.md
```diff
@@ -133,6 +133,115 @@ HTML markup should never be used outside of code examples, though Markdown can b
...
+Only a trailing standalone line comment counts. `// Outputs:` inside a string, a block comment, after code on the same line, or followed by more code is treated as code. The older ` ```expected-output ` fence still parses; do not use both.
+
+##### Hidden setup
+
+If the example needs a helper function, a post, or an option, put that in a setup Blueprint ([Playground Blueprint](https://developer.wordpress.org/playground/blueprints/) JSON). It runs first and never appears on the page.
+
+Inline, directly before or after the PHP fence:
+````
+```setup-blueprint
+{ "steps": [ { "step": "writeFile", "path": "/wordpress/wp-content/mu-plugins/docs-fixture.php", "data": "<?php\nfunction docs_fixture_greeting() { return 'Hello'; }\n" } ] }
```
jrfnl, line 209: This is an unrealistic example as the function definition could just be in the PHP code block. I believe this documentation would be better served if the code sample would be more realistic, i.e. show something which actually needs some sort of real setup.

inline-documentation-standards/php.md
```diff
@@ -133,6 +133,115 @@ HTML markup should never be used outside of code examples, though Markdown can b
...
+```setup-blueprint shared-greeting
+{ "steps": [ ... ] }
+```
+```php interactive setup-blueprint=shared-greeting
+echo docs_shared_greeting( 'first' );
+// Outputs: Hello, first
+```
+````
+
+Names are case-sensitive. `blueprint=` and `setupblueprint=` are not recognized.
```
jrfnl, line 228: These phrases are confusing and look unrelated.

The first phrase talks about case-sensitivity, but the examples are not related to case-sensitivity. The examples also are about the key, not the name `shared-greeting`.

This needs work.

inline-documentation-standards/php.md
```diff
@@ -133,6 +133,115 @@ HTML markup should never be used outside of code examples, though Markdown can b
...
+
+##### Before you commit
+
+- Fence reads exactly `php interactive`.
+- No `<?php`, no `require`.
+- Prints something, and the `// Outputs` comment matches.
+- Still readable as plain text in the source. If not, keep the indented version.
```
jrfnl, line 235:
```suggestion
Checklist:
- [ ] Fence reads exactly `php interactive`.
- [ ] PHP code sample contains no `<?php`, no `require`.
- [ ] The code sample prints something, and the `// Outputs` comment matches the expected output.
- [ ] The DocBlock with the example is still readable as plain text in the source. If not, keep the indented version of the example code.
```

(Change to checklist only if the `- [ ]` syntax is supported. I'm not sure that it is.)

inline-documentation-standards/php.md
```diff
@@ -133,6 +133,115 @@ HTML markup should never be used outside of code examples, though Markdown can b
...
+
+To test, paste the code into a `<php-snippet>` on any HTML page per the [Playground guide](https://developer.wordpress.org/playground/handbook/guides/php-code-snippets/).
+
+##### Timing and feedback
+
+The Code Reference is built from the current stable release, so an example in `trunk` appears when that release ships.
```
jrfnl, line 241:
```suggestion
The Code Reference is built from the current stable release, so an example added in `trunk` only appears when that WordPress release ships.
```

jrfnl (changes requested):
As a general note/question - this is quite a large chunk of text just about runnable examples and distracts a lot from the more generic PHP documentation standard the page is about.

Should this topic possibly get its own page ? which would then be linked from point 2 in the "Description" section ?

I'm also thinking about the potential that runnable code examples will at some point also be supported for JavaScript and by having a separate page about runnable code examples, that could all be documented in one place instead of having to duplicate a lot between the PHP and JS documentation standards.

Think it over if you will.

inline-documentation-standards/php.md
zzap: +1 for "End with **an** `// Outputs` comment."

zzap (changes requested):
First, I believe the chosen placement is wrong. Reading from the top of the page and arriving at `Runnable examples` before `@since`, `@param`, and other official DocBlock keywords, which are considered mandatory for any proper documentation, feels very much "out of the blue". 

Second, the amount of detailed info in this introductory part of the page is unbalanced and at odds with its importance compared to the surroundings.

If we recommend NOT using it all the time, and if it is "opt-in", then we can not, under any circumstances, place it above mandatory parts. And especially, we can not explain it in full detail in such a placement.

My suggestion is: 
- First mention it in the `Other Descriptions` section, with one or two sentences, explaining that it is possible to include interactive examples for code that needs more explanation.  
- Then add it as a list item under `Functions & Class Methods`, again saying it's opt-in, and include just a skeleton in the code example (only opening and closing lines). 
- Then add complete explanation as 1.3 Runnable examples (between 1.2 Deprecated functions and 2. Classes).

As for the content itself, some sentences are unclear and I will add comments in their respective threads, but what I can suggest just from scanning is to give a proper name to things that are new. 

For example, "an `// Outputs` comment" needs to have an official term . Very much like `#!` is called `shebang`. Using **"an `// Outputs` comment"** in written form is, at best, clunky. But try saying it verbaly (in a talk, workshop, Contributor Day, conversation..) It's going to be very difficult to talk about it in such format. So, define its name now, explain it early in Runnable examples documentation, and use it consistently afterwards.

inline-documentation-standards/php.md
```diff
@@ -133,6 +133,115 @@ HTML markup should never be used outside of code examples, though Markdown can b
      * @link https://core.trac.wordpress.org/ticket/20000
     ```
 
+#### Runnable examples
+
+A DocBlock example can be marked as runnable. The Code Reference then shows it with a **Run** button, and readers execute and edit it in the browser through [WordPress Playground](https://developer.wordpress.org/playground/handbook/guides/php-code-snippets/).
```
zzap, line 138: A code example in DocBlock can become interactive by marking it with `interactive` keyword. On the Code Reference page, visitors can edit and execute the code via "Run" button. This is powered by  [WordPress Playground](https://developer.wordpress.org/playground/handbook/guides/php-code-snippets/).

inline-documentation-standards/php.md
zzap: Is `indented` important here? I understand the point of this paragraph is:

```
This is opt-in. Existing examples keep working as intended. Consider converting them to interactive examples only when running the code teaches something that simply reading it cannot.
```

Does indentation change this behaviour?

zzap: I got my answer afterwards. I don't think we need to mention indentation here.

inline-documentation-standards/php.md
zzap, line 169: I had to read this 3 times to understand. I think the list would be clearer:

> Do **not** include:
>- `<?php` - it's generated automatically 
>- `require '/wordpress/wp-load.php';` - WordPress is already loaded
> 
> Including any of these will disable the interactive behaviour of the example.

inline-documentation-standards/php.md
zzap, line 175: Getting back to my remark about naming "a `// Outputs` comment". 

To me it sounds like `Output block`.

inline-documentation-standards/php.md
```diff
@@ -133,6 +133,115 @@ HTML markup should never be used outside of code examples, though Markdown can b
...
+
+End with a `// Outputs` comment. The parser strips it from the code that runs, shows it as the expected result before the reader selects **Run**, and keeps it for later output checks.
+
+One line — everything after the colon is literal:
+```php
+echo esc_html( '<egg>' );
+// Outputs: <egg>
+```
+
+Several lines — empty `// Outputs:` starts a block, each `//` line is one output line, a final empty `//` keeps a trailing newline:
```
zzap, line 183: Let's change `Several lines` to `Multiline`. It aligns better with the rest of the DocBlock vocabulary.

zzap: Also, I would add a sentence in the intro of this section: 

> The Output block can be a single-line or a multiline example.

inline-documentation-standards/php.md
zzap, line 175: I think we don't need to know what the parser is doing. I would just say:

>The Output block is displayed as the output of the initial interactive example, before the reader executes or modifies it.

inline-documentation-standards/php.md
zzap, line 200: I would add examples of what is wrong and what is correct here. As the rest of the coding standards docs do.

inline-documentation-standards/php.md
```diff
@@ -133,6 +133,115 @@ HTML markup should never be used outside of code examples, though Markdown can b
...
+
+JSON-encoded — for trailing spaces, tabs, or repeated newlines. Must be one JSON string:
+```php
+echo "done ";
+// Outputs (JSON-encoded): "done "
+```
+
+Only a trailing standalone line comment counts. `// Outputs:` inside a string, a block comment, after code on the same line, or followed by more code is treated as code. The older ` ```expected-output ` fence still parses; do not use both.
+
+##### Hidden setup
```
zzap, line 202: This whole section, in my opinion, should live elsewhere, and here we can link to it.

inline-documentation-standards/php.md
```diff
@@ -133,6 +133,115 @@ HTML markup should never be used outside of code examples, though Markdown can b
...
+##### Before you commit
+
+- Fence reads exactly `php interactive`.
+- No `<?php`, no `require`.
+- Prints something, and the `// Outputs` comment matches.
+- Still readable as plain text in the source. If not, keep the indented version.
+
+To test, paste the code into a `<php-snippet>` on any HTML page per the [Playground guide](https://developer.wordpress.org/playground/handbook/guides/php-code-snippets/).
+
+##### Timing and feedback
```
zzap, line 239: This section is not for the intended audience. We are mixing user and contributor documentation here, and I would suggest removing this part altogether.

inline-documentation-standards/php.md
jrfnl: @zzap "Code being rendered as code by making it an indented block" is a markdown feature. Backtick fencing is an alternative to using indentation. With that in mind, I find the phrasing "indented examples" clear, but I also realize while writing this, that that presumes the person reading it actually knows markdown syntax and knows that the PHP docs are converted to markdown, so I can imagine the phrasing could do with some additional clarification.

zzap: Yes, we need to be clear with who we are talking to, and what is their "assumed knowledge" to understand and use this documentation.

inline-documentation-standards/php.md
GaryJones: I'd go with an alternative:

> End with an output comment (`// Outputs: ...`).

Separates outs the English description from the required code comment format, and disambiguates the choice of indefinite article.

